PR #198, opened 2021-08-26
pools: Introduce a cap in terms of TVL for a pool

axic:


axic:
Can we close this? I don't think this will ever be implemented.

PR #230, opened 2021-08-28
token: Change ERC20OwnerMintableToken to use Ownable

axic:


mijovic:
Closing for now, we never went this way, and I think we will not need tokens with transferable ownershit

PR #338, opened 2021-10-04
controller: Adding liquidity with correct ratio.

mijovic:


contracts/TempusController.sol
```diff
@@ -308,6 +318,33 @@ contract TempusController is PermanentlyOwnable, ReentrancyGuard {
         }
     }
 
+    function _provideLiquidity(ITempusAMM tempusAMM, uint256 sharesAmount) private {
```
yuval-weiss, line 321: can we consume `_provideLiquidity` in `_depositAndProvideLiquidity`?

mijovic: Yes we can, thanks for raising :) Will update now

contracts/TempusController.sol
```diff
@@ -275,17 +292,49 @@ contract TempusController is PermanentlyOwnable, ReentrancyGuard {
...
+            vault,
+            poolId,
+            ammTokens,
+            ammBalances,
+            mintedShares,
+            msg.sender
+        );
 
+        // Send remaining Shares to user
+        if (sharesUsed[0] < mintedShares) {
```
mijovic, line 308: Instead of this we can check if some balance is left in controller instead, let me know what do you think @yuval-weiss

contracts/TempusController.sol
```diff
@@ -275,17 +292,49 @@ contract TempusController is PermanentlyOwnable, ReentrancyGuard {
...
         uint256[] memory ammLiquidityProvisionAmounts = new uint256[](2);
 
         (ammLiquidityProvisionAmounts[0], ammLiquidityProvisionAmounts[1]) = (
-            ammTokens[0].balanceOf(address(this)).mulf18(ammDepositPercentages[0]),
-            ammTokens[1].balanceOf(address(this)).mulf18(ammDepositPercentages[1])
+            sharesAmount.mulf18(ammDepositPercentages[0]),
+            sharesAmount.mulf18(ammDepositPercentages[1])
         );
 
+        if (sender != address(this)) {
```
mijovic, line 333: If sender is controller, funds are already here

PR #358, opened 2021-10-17
Fix return names code arena 4

mijovic:


contracts/amm/TempusAMM.sol
```diff
@@ -529,8 +529,6 @@ contract TempusAMM is BaseGeneralPool, BaseMinimalSwapInfoPool, StableMath, IRat
 
         amountsOut.div(tokenRates, _TEMPUS_SHARE_PRECISION);
         dueProtocolFeeAmounts.div(tokenRates, _TEMPUS_SHARE_PRECISION);
-
-        return (bptAmountIn, amountsOut, dueProtocolFeeAmounts);
```
mijovic, line 533: Coverage decreased because we have deleted covered line

PR #368, opened 2021-10-20
pool: Refactor matured logic

mijovic:
This is also gas saving as we do not have `SLOAD` in checking if the pool matured.
Also, interest rate on maturity will be set on first action after maturity


test/TempusPool.Redeem.test.ts
```diff
@@ -272,6 +276,8 @@ describeForEachPool("TempusPool Redeem", (pool:ITestPool) =>
     
         await pool.setInterestRate(2.5);
         await pool.fastForwardToMaturity();
+        // just to update maturity interest rate
+        await pool.redeemToYBT(owner, 0.1, 0.1);
```
RedFox20, line 280: @mijovic I am worried about this. We have a dependency on someone doing a redeem before the maturity interest rate is in effect. It really doesn't have many side-effects, it's just an annoying state for the contract to be in...

test/TempusPool.Redeem.test.ts
```diff
@@ -219,13 +219,15 @@ describeForEachPool("TempusPool Redeem", (pool:ITestPool) =>
...
 
     await pool.depositYBT(user, 100, /*recipient:*/user);
+    await pool.depositYBT(owner, 10);
     (await pool.userState(user)).expect(100, 100, /*yieldBearing:*/100);
 
     await pool.setInterestRate(2.0);
     if (pool.yieldPeggedToAsset)
     {
         (await pool.userState(user)).expect(100, 100, /*yieldBearing:*/200);
         await pool.fastForwardToMaturity();
```
RedFox20, line 229: Maybe we should add something to `fastForwardToMaturity` instead?

test/TempusPool.Redeem.test.ts
RedFox20: Yes, it was essentially the same situation. From tests standpoint, we just added a bunch of confusing redeemToYBT calls. Discussed this with @mijovic on Discord and `fastForwardToMaturity()` should handle finalization.

PR #374, opened 2021-10-21
Fixing Typos

mijovic:


mijovic:
Unused importes went in somehow, will fix PR now with that

PR #400, opened 2021-10-28
vesting: Allow retrieving the earned tokens during stopVesting

axic:


contracts/token/vesting/IERC20Vesting.sol
```diff
@@ -65,6 +65,8 @@ interface IERC20Vesting {
     function transferVesting(address oldAddress, address newAddress) external;
 
     /// @dev Stops vesting for receiver and sends all tokens back to wallet
+    /// Any earned claimable amount is still claimable through `claim()`.
```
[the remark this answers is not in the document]
mijovic: I like using these, not sure if it is useful at all...

contracts/token/vesting/IERC20Vesting.sol
```diff
@@ -65,6 +65,8 @@ interface IERC20Vesting {
...
+    /// Please note however that an account cannot be used again.
```
mijovic, line 69:
```suggestion
    /// @notice Account cannot be used again as the vesting contract receiver.
```

contracts/token/vesting/IERC20Vesting.sol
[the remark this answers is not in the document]
axic: `@notice` vs `@dev` is misleading, see https://docs.soliditylang.org/en/v0.8.9/natspec-format.html?highlight=%40notice#tags

contracts/token/vesting/ERC20Vesting.sol
```diff
@@ -86,13 +86,20 @@ contract ERC20Vesting is IERC20Vesting {
     function stopVesting(address receiver) external override onlyWallet {
         require(receiver != address(0), "Receiver cannot be 0.");
 
-        VestingTerms memory terms = vestingTerms[receiver];
+        VestingTerms storage terms = vestingTerms[receiver];
```
axic, line 89: This is the only "questionable" part. Using storage reference here to have the terms updated below. Alternatively could just explicitly put in the storage access.

mijovic: I don't mind it at all, like for me it is even better like this as it we reuse it in many different places

mijovic: I think we should use it as you did here

contracts/token/vesting/IERC20Vesting.sol
mijovic: Can we add that account can't be used in vesting contract as receiver, `account cannot be used again` sound misleading

axic: Done.

mijovic (approved):

PR #408, opened 2021-10-29
vesting: Allow claiming maximum claimable

axic:


test/token/vesting/ERC20TokenVesting.test.ts
```diff
@@ -212,19 +212,68 @@ describe("ERC20 Vesting", async () => {
...
+        {startTime: startTime, period: period, amount: 30, claimed: 0}
+      );
+
+      await increaseTime(period / 2);
+
+      await vesting.claim(user, 7);
+      expect(+await token.balanceOf(user)).to.equal(7);
+
+      await vesting.claim(user, 7);
+      expect(+await token.balanceOf(user)).to.equal(14);
```
axic, line 272: Hardcoded specific amounts not using `claimable` here just made this work properly.

contracts/token/vesting/IERC20Vesting.sol
```diff
@@ -50,7 +50,10 @@ interface IERC20Vesting {
     /// @param terms Vesting terms for all accounts
     function startVestingBatch(address[] calldata receivers, VestingTerms[] calldata terms) external;
 
-    /// @dev Transfers tokens to a given address
+    /// @dev Transfers all vested tokens to the sender
+    function claim() external;
+
+    /// @dev Transfers a part of vested tokens to the sender
```
mijovic, line 56: We should add a comment to state that passing `type(uint256).max` will claim all vested tokens.

axic: Done!

mijovic (approved):

PR #411, opened 2021-11-01
vesting: Ensure claimable() only returns for valid schedules

axic:


contracts/token/vesting/ERC20Vesting.sol
```diff
@@ -99,7 +99,9 @@ contract ERC20Vesting is IERC20Vesting {
     }
 
     function claimable(address receiver) external view override returns (uint256) {
-        return _claimable(vestingTerms[receiver]);
+        VestingTerms memory terms = vestingTerms[receiver];
+        require(isScheduleValid(terms), "No vesting data for receiver.");
```
yuval-weiss, line 103: Can we move the require statement to the internal `_claimable` function? The other consumer of `_claimable` (`claim()`) also checks `isScheduleValid`...

axic:
@mijovic this is still a question.

mijovic:
> @mijovic this is still a question.

Well if we want checks to be in place, we should try including this PR as well.
Maybe moving check to _claimable as @yuval-weiss suggested?

axic:
> Maybe moving check to `_claimable` as @yuval-weiss suggested?

I'm not sure actually, because most of the functions need this check, even those which are not using `_claimable`. So in that sense perhaps it is more clear to have it in the function body as opposed to hidden in a helper.

contracts/token/vesting/ERC20Vesting.sol
```diff
@@ -113,7 +113,9 @@ contract ERC20Vesting is IERC20Vesting {
     }
 
     function claimable(address receiver) external view override returns (uint256) {
```
axic, line 115: Alternative is if we just add a note to natspec saying this will return 0 for non-existent accounts.

mijovic: this is also fine by me

mijovic (approved):

PR #412, opened 2021-11-01
vesting: Adding few more tests.

mijovic:


test/token/vesting/ERC20TokenVesting.test.ts
```diff
@@ -1,5 +1,5 @@
 import { ethers } from "hardhat";
-import { expect } from "chai";
+import { expect, use } from "chai";
```
axic, line 2: What is `use` used for here?

test/token/vesting/ERC20TokenVesting.test.ts
```diff
@@ -143,6 +143,16 @@ describe("ERC20 Vesting", async () => {
       expect(await token.balanceOf(owner)).to.equal(270);
     });
 
+    it("Expected claimable tokes is 0  after startVesting with startTime in future", async () => {
```
axic, line 146:
```suggestion
    it("Expected claimable tokens is 0 after startVesting with startTime in future", async () => {
```

test/token/vesting/ERC20TokenVesting.test.ts
mijovic: ah, probably it was added as I accidentally added use instead of user somewhere. Removing now

axic (approved):

PR #414, opened 2021-11-02
utils: Remove Context from Ownable

axic:


contracts/utils/Ownable.sol
```diff
@@ -46,7 +46,7 @@ abstract contract Ownable is Context {
      * Can only be called by the proposed owner.
      */
     function acceptOwnership() public virtual {
-        require(msg.sender == _proposedOwner, "Ownable: Only proposed owner can accept ownership");
+        require(_msgSender() == _proposedOwner, "Ownable: Only proposed owner can accept ownership");
```
axic, line 49: Since this is using `Context`, this is required -- but why is this using `Context`?

axic: Can we remove `Context`?

mijovic: I was thinking the same. This was just copy/paste from OZ. Let's remove it please :)

mijovic (approved):

PR #416, opened 2021-11-02
vesting: Use SafeERC20

axic:


contracts/token/vesting/ERC20Vesting.sol
```diff
@@ -34,7 +38,7 @@ contract ERC20Vesting is IERC20Vesting {
         require(!isScheduleValid(vestingTerms[receiver]), "Vesting already started for account.");
 
         vestingTerms[receiver] = terms;
-        token.transferFrom(wallet, address(this), terms.amount);
+        token.safeTransferFrom(wallet, address(this), terms.amount);
```
axic, line 41: Even if not using `SafeEC20`, the return value should be checked for strict ERC-20 compliance.

mijovic: It makes this more general, so I would say it is better to use SafeERC20 anyway.

mijovic (approved):

mijovic (approved):

PR #417, opened 2021-11-02
vesting: Ensure that vesting cannot be set up with invalid token/wallet

axic:


contracts/token/vesting/ERC20Vesting.sol
```diff
@@ -20,6 +20,9 @@ contract ERC20Vesting is IERC20Vesting {
     /// @param _token Address of ERC20 token associated with this vesting contract
     /// @param _wallet Address of account that starts and stops vesting for different parties
     constructor(IERC20 _token, address _wallet) {
+        require(address(_token) != address(0), "Token cannot be 0.");
+        require(address(_wallet) != address(0), "Wallet cannot be 0.");
```
mijovic, line 24: We will need tests for these? If it is too much fuzz, I completely agree with leaving this untested

test/token/vesting/ERC20TokenVesting.test.ts
```diff
@@ -26,6 +26,17 @@ describe("ERC20 Vesting", async () => {
   });
 
   describe("Invalid inputs", async () => {
+    it("Constructor reverts on input", async () => {
+      (await expectRevert(ERC20Vesting.create(
+        "0x0000000000000000000000000000000000000000",
```
axic, line 31: This needs to be casting into the testing type `ERC20` 😬

mijovic (approved):

PR #420, opened 2021-11-02
ownable: Emit event for proposal of transfer

axic:


mijovic (approved):
Very simple and clear. Good to go

PR #421, opened 2021-11-04
pool: Adding TokenData helper.

mijovic:


contracts/TempusPool.sol
```diff
@@ -11,6 +11,12 @@ import "./token/YieldShare.sol";
 import "./math/Fixed256xVar.sol";
 import "./utils/UntrustedERC20.sol";
 
+/// @dev helper struct to store name and symbol for the token
+struct TokenData {
```
mijovic, line 15: Not sure about name for this struct

contracts/mocks/aave/AavePoolMock.sol
```diff
@@ -24,11 +24,11 @@ contract AavePoolMock is ILendingPool {
         IERC20 asset,
         uint128 index,
         uint8 decimals,
-        string memory aTokenName,
+        string memory aTokenData,
```
mijovic, line 27: find and replace issue, will revert it now

contracts/TempusPool.sol
```diff
@@ -58,10 +64,8 @@ abstract contract TempusPool is ITempusPool {
     /// @param initInterestRate initial interest rate of the pool
     /// @param exchangeRateOne 1.0 expressed in exchange rate decimal precision
     /// @param estimatedFinalYield estimated yield for the whole lifetime of the pool
-    /// @param principalName name of Tempus Principal Share
-    /// @param principalSymbol symbol of Tempus Principal Share
-    /// @param yieldName name of Tempus Yield Share
-    /// @param yieldSymbol symbol of Tempus Yield Share
+    /// @param principalsData Tempus Principals name and symbol
+    /// @param yieldsData name of Tempus Yields name and symbol
```
axic, line 68:
```suggestion
    /// @param yieldsData Tempus Yields name and symbol
```

contracts/TempusPool.sol
axic: What is the goal of this change?

mijovic: Adding additional param raises stack to deep error, so I did this in separate PR to have 1 change at a time

axic: Where does it raise stack too deep error?

axic: Also perhaps `TokenSettings` / `TokenProperties` may be a better name?

axic (approved):
Otherwise looks good.

contracts/TempusPool.sol
axic: > Where does it raise stack too deep error?

Ah it raises in #422.

PR #422, opened 2021-11-04
pool: Add owner as constructor param

mijovic:


contracts/TempusPool.sol
```diff
@@ -72,6 +73,7 @@ abstract contract TempusPool is ITempusPool {
         address _yieldBearingToken,
         address _backingToken,
         address ctrl,
+        address _owner,
```
axic, line 76: Hm, it would be nice to just get this deployed via the controller, but that runs into the factory question.

contracts/TempusPool.sol
```diff
@@ -72,6 +73,7 @@ abstract contract TempusPool is ITempusPool {
         address _yieldBearingToken,
         address _backingToken,
         address ctrl,
```
axic, line 75: It would maybe make sense moving controller/owner to be the first items (later).

axic (changes requested):
Hmm, I think there should be at least one test checking the ownership of a deployed pool. It seems no test was broken by this change.

test/utils/TempusPool.ts
```diff
@@ -190,6 +190,7 @@ export class TempusPool extends ContractBase {
         owner,
         yieldToken.address,
         controller.address,
+        owner.address,
```
yuval-weiss, line 193: The deploy functions' `owner` param in the current implementation of TempusPool.ts (before this PR) is actually the deployer of the contract, which is also the owner in the current `master` implementation. This PR separates the contract deployer from the `owner` by making `owner` an actual constructor param - so I think we should separate the `deploy` functions' `owner` param into 2 parameters - `deployer` (which will be used to deploy the contract) and `owner` (which will be passed to the TempusPool constructor).

After we add that I suggest we add a unit test in `TempusPool.Deploy.test` the verifies that deploying a pool with a `deployer` different than the provided `owner` value sets the `TempusPool.owner` property properly.

contracts/TempusPool.sol
```diff
@@ -82,13 +84,14 @@ abstract contract TempusPool is ITempusPool {
     ) {
         require(maturity > block.timestamp, "maturityTime is after startTime");
         require(ctrl != address(0), "controller can not be zero");
+        require(_owner != address(0), "owner can not be zero");
         require(initInterestRate > 0, "initInterestRate can not be zero");
```
yuval-weiss, line 88: another possible `require` statement that we can add here is `require(_owner.balance > 0);`. It's not super critical but I think making sure the provided `owner` address is funded with some eth is a nice sanity check that we didn't accidentally have a typo or that we copied the wrong address from testnet or something.

contracts/TempusPool.sol
mijovic: I would not start doing factory now, maybe for v2 it would make sense

contracts/TempusPool.sol
mijovic: Added the check

contracts/TempusPool.sol
mijovic: Moved them to be the first args

contracts/TempusPool.sol
```diff
@@ -82,13 +84,15 @@ abstract contract TempusPool is ITempusPool {
     ) {
         require(maturity > block.timestamp, "maturityTime is after startTime");
         require(ctrl != address(0), "controller can not be zero");
+        require(_owner != address(0), "owner can not be zero");
+        require(_owner.balance > 0, "owner balance can not be zero");
```
axic, line 88: Why?

test/utils/TempusPool.ts
```diff
@@ -88,6 +88,7 @@ export class TempusPool extends ContractBase {
    * @param tempusShareNames Symbol names for TPS+TYS
    */
   static async deployAave(
+    deployer:Signer,
```
axic, line 91: Isn't this more clear?
```suggestion
    controller:Signer,
```

PR #423, opened 2021-11-05
pool: Exceptionally stop pool in case of a long negative yield period

axic:
This is mostly here for a discussion starter.

### DIFF
--- a/contracts/ITempusPool.sol
+++ b/contracts/ITempusPool.sol
@@ -77,7 +77,16 @@ interface ITempusPool is ITempusFees {
     /// @return Maturity time of the pool.
     function maturityTime() external view returns (uint256);
 
+    /// @return Time of exceptional halting of the pool.
+    /// In case the pool is still in operation, this must return type(uint256).max.
+    function exceptionalHaltTime() external view returns (uint256);
+
+    /// @return The maximum allowed time (in seconds) to pass with negative yield.
+    function maximumNegativeYieldDuration() external view returns (uint256);
+
     /// @return True if maturity has been reached and the pool was finalized.
+    ///         This also includes the case when maturity was triggered due to
+    ///         exceptional conditions (negative yield periods).
     function matured() external view returns (bool);
 
     /// Finalizes the pool. This can only happen on or after `maturityTime`.
--- a/contracts/TempusPool.sol
+++ b/contracts/TempusPool.sol
@@ -28,11 +28,14 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable {
 
     uint256 public constant override version = 1;
 
+    uint256 public constant override maximumNegativeYieldDuration = 7 days;
+
     address public immutable override yieldBearingToken;
     address public immutable override backingToken;
 
     uint256 public immutable override startTime;
     uint256 public immutable override maturityTime;
+    uint256 public override exceptionalHaltTime = type(uint256).max;
 
     uint256 public immutable override initialInterestRate;
     uint256 public override maturityInterestRate;
@@ -53,6 +56,9 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable {
     uint256 public immutable override maxEarlyRedeemFee;
     uint256 public immutable override maxMatureRedeemFee;
     uint256 public override totalFees;
+
+    /// Timestamp when the negative yield period was entered.
+    uint256 private negativeYieldStartTime;
 
     /// Constructs Pool with underlying token, start and maturity date
     /// @param _yieldBearingToken Yield Bearing Token, such as cDAI or aUSDC
@@ -116,7 +122,7 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable {
         returns (uint256 backingTokenAmount);
 
     function matured() public view override returns (bool) {
-        return block.timestamp >= maturityTime;
+        return (block.timestamp >= maturityTime) || (block.timestamp >= exceptionalHaltTime);
     }
 
     function getFeesConfig() external view override returns (FeesConfig memory) {
@@ -185,7 +191,7 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable {
             uint256 rate
         )
     {
-        rate = updateInterestRate();
+        rate = updateAndValidateInterestRate();
 
         require(!matured(), "Maturity reached.");
         require(rate >= initialInterestRate, "Negative yield!");
@@ -271,7 +277,7 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable {
         require(IERC20(address(principalShare)).balanceOf(from) >= principalAmount, "Insufficient principals.");
         require(IERC20(address(yieldShare)).balanceOf(from) >= yieldAmount, "Insufficient yields.");
 
-        uint256 currentRate = updateInterestRate();
+        uint256 currentRate = updateAndValidateInterestRate();
 
         if (matured()) {
             finalize();
@@ -445,6 +451,33 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable {
         uint256 currentRate = currentInterestRate();
         (uint256 yieldTokens, uint256 backingTokens, , ) = getRedemptionAmounts(principals, yields, currentRate);
         return toBackingToken ? backingTokens : yieldTokens;
+    }
+
+    /// @dev This updates the underlying pool's interest rate, and also updates the internal
+    ///      tracking of negative interest periods.
+    function updateAndValidateInterestRate() internal returns (uint256 rate) {
+        rate = updateInterestRate();
+
+        // Short circuit. No need for the below after maturity.
+        if (matured()) {
+            return rate;
+        }
+
+        if (rate < initialInterestRate) {
+            if (negativeYieldStartTime == 0) {
+                // Entering a negative yield period.
+                negativeYieldStartTime = block.timestamp;
+            } else {
+                // Already in a negative yield period, exceeding the duration.
+                if ((negativeYieldStartTime + maximumNegativeYieldDuration) <= block.timestamp) {
+                    exceptionalHaltTime = block.timestamp;
+                    assert(matured());
+                }
+            }
+        } else {
+            // Reset period.
+            negativeYieldStartTime = 0;
+        }
     }
 
     /// @dev This updates the underlying pool's interest rate
--- a/test/TempusPool.Redeem.test.ts
+++ b/test/TempusPool.Redeem.test.ts
@@ -176,6 +176,11 @@ describeForEachPool("TempusPool Redeem", (pool:PoolTestFixture) =>
     await pool.setTimeDaysAfterPoolStart(15);
     await pool.setInterestRate(0.1);
     await redeemAndCheckYBT(user, { redeem: { tps: 70, tys: 70 }, balanceAfter: { tps: 0, tys: 0 }, ybtAfter: { pegged: 20, unpegged: 200 } });
+
+    // Check that the pool HAS matured/halted yet.
+    expect(await pool.tempus.matured()).to.equal(true);
+    // TODO: extract expected timestamp from the helpers above
+    expect(await pool.tempus.exceptionalHaltTime()).to.not.equal(null);
   });
 
   it.includeIntegration("Should work before maturity with equal shares, with fluctuating yield", async () =>
@@ -206,6 +211,10 @@ describeForEachPool("TempusPool Redeem", (pool:PoolTestFixture) =>
     await pool.setTimeDaysAfterPoolStart(14);
     await pool.setInterestRate(1);
     await redeemAndCheckYBT(user, { redeem: { tps: 60, tys: 60 }, balanceAfter: { tps: 0, tys: 0 }, ybtAfter: { pegged: 200, unpegged: 200 } });
+
+    // Check that the pool HAS NOT matured/halted yet.
+    expect(await pool.tempus.matured()).to.equal(false);
+    expect(await pool.tempus.exceptionalHaltTime()).to.equal(null);
   });
 
   it.includeIntegration("Should work after maturity with negative yield", async () =>
--- a/test/TempusPool.Redeem.test.ts
+++ b/test/TempusPool.Redeem.test.ts
@@ -192,8 +192,8 @@ describeForEachPool("TempusPool Redeem", (pool:PoolTestFixture) =>
     await redeemAndCheckYBT(user, { redeem: { tps: 10, tys: 10 }, balanceAfter: { tps: 90, tys: 90 }, ybtAfter: { pegged: 99, unpegged: 110 } });
 
     await pool.setTimeDaysAfterPoolStart(5);
-    await pool.setInterestRate(1);
-    await redeemAndCheckYBT(user, { redeem: { tps: 10, tys: 10 }, balanceAfter: { tps: 80, tys: 80 }, ybtAfter: { pegged: 120, unpegged: 120 } });
+    await pool.setInterestRate(1.1);
+    await redeemAndCheckYBT(user, { redeem: { tps: 10, tys: 10 }, balanceAfter: { tps: 80, tys: 80 }, ybtAfter: { pegged: 132, unpegged: 120 } });
 
     await pool.setTimeDaysAfterPoolStart(8);
     await pool.setInterestRate(0.6);
--- a/test/utils/TempusPool.ts
+++ b/test/utils/TempusPool.ts
@@ -1,5 +1,5 @@
 import { BigNumber, BytesLike, Contract, Transaction } from "ethers";
-import { NumberOrString, toWei, parseDecimal, formatDecimal } from "./Decimal";
+import { NumberOrString, toWei, parseDecimal, formatDecimal, MAX_UINT256 } from "./Decimal";
 import { ContractBase, Signer, SignerOrAddress, addressOf } from "./ContractBase";
 import { ERC20 } from "./ERC20";
 import { IERC20 } from "./IERC20";
@@ -395,6 +395,26 @@ export class TempusPool extends ContractBase {
   async maturityTime(): Promise<NumberOrString> {
     let maturity:BigNumber = await this.contract.maturityTime();
     return maturity.toNumber();
+  }
+
+  /**
+   * @returns The exceptional halt time of the pool
+   * @note This returns null in case it is not set (i.e. has the special value of `type(uin256).max`)
+   */
+  async exceptionalHaltTime(): Promise<NumberOrString | null> {
+    let exceptionalHaltTime:BigNumber = await this.contract.exceptionalHaltTime();
+    if (exceptionalHaltTime.toHexString() === MAX_UINT256) {
+      return null;
+    }
+    return exceptionalHaltTime.toNumber();
+  }
+
+  /**
+   * @returns The maximum allowed duration of negative yield periods (in seconds)
+   */
+  async maximumNegativeYieldDuration(): Promise<NumberOrString> {
+    let maximumNegativeYieldDuration:BigNumber = await this.contract.maximumNegativeYieldDuration();
+    return maximumNegativeYieldDuration.toNumber();
   }
 
   /**
--- a/test/TempusPool.Deploy.test.ts
+++ b/test/TempusPool.Deploy.test.ts
@@ -31,9 +31,11 @@ describeForEachPool("TempusPool Deploy", (testPool:PoolTestFixture) =>
     expect(await pool.maturityTime()).to.equal(testPool.maturityTime);
   });
 
-  it("Maturity should not be set", async () =>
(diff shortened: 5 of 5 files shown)


contracts/TempusPool.sol
```diff
@@ -24,13 +24,16 @@ abstract contract TempusPool is ITempusPool {
...
+    uint256 public constant override maximumNegativeYieldDuration = 7 days;
+
+    uint256 public constant override version = 1;
 
     address public immutable override yieldBearingToken;
     address public immutable override backingToken;
 
     uint256 public immutable override startTime;
     uint256 public immutable override maturityTime;
+    uint256 public override exceptionalHaltTime = type(uint256).max;
```
axic, line 36: This is a trick so that we don't need to check `!= 0` in `halted` below.

contracts/TempusPool.sol
```diff
@@ -127,6 +133,10 @@ abstract contract TempusPool is ITempusPool {
         return block.timestamp >= maturityTime;
     }
 
+    function halted() public view override returns (bool) {
```
axic, line 136: Not sure we need this helper -- we could incorporate this into `matured`.

contracts/TempusPool.sol
```diff
@@ -454,6 +465,27 @@ abstract contract TempusPool is ITempusPool {
         return toBackingToken ? backingTokens : yieldTokens;
     }
 
+    /// @dev This updates the underlying pool's interest rate, and also updates the internal
+    ///      tracking of negative interest periods.
+    function updateAndCheckInterestRate() internal returns (uint256 rate) {
```
axic, line 470: I wonder if this helper is the right approach, or doing some modifier/helper function and explicitly calling it on the entry points, or on `burnShares`/`mintShares`.

contracts/TempusPool.sol
```diff
@@ -454,6 +465,27 @@ abstract contract TempusPool is ITempusPool {
...
+                lastNegativeYield = block.timestamp;
+            } else {
+                // Already in a negative yield period, exceeding the duration.
+                if ((lastNegativeYield + maximumNegativeYieldDuration) >= block.timestamp) {
+                    exceptionalHaltTime = block.timestamp;
+                }
+            }
+        } else {
+            // Reset period.
+            lastNegativeYield = 0;
```
axic, line 485: This here in the generic case means resetting the storage value with 0.  This adds a constant 100 gas overhead.

Any better ideas?

contracts/TempusPool.sol
```diff
@@ -47,15 +50,18 @@ abstract contract TempusPool is ITempusPool {
...
     uint256 public immutable override maxDepositFee;
     uint256 public immutable override maxEarlyRedeemFee;
     uint256 public immutable override maxMatureRedeemFee;
 
-    /// total amount of fees accumulated in pool
+    /// Total amount of fees accumulated in pool.
     uint256 public override totalFees;
 
+    /// Timestamp when the negative yield period was entered.
+    uint256 private lastNegativeYield;
```
axic, line 63: Potentially could just rework this to merge `exceptionalHaltTime` and `lastNegativeYield` into a single variable.  In this case though we definitely need to check for `!= 0` in the matured/halted helpers.

contracts/TempusPool.sol
axic: Did that in a second commit.

contracts/TempusPool.sol
```diff
@@ -24,13 +24,16 @@ abstract contract TempusPool is ITempusPool {
     using UntrustedERC20 for IERC20;
     using Fixed256xVar for uint256;
 
-    uint public constant override version = 1;
+    uint256 public constant override maximumNegativeYieldDuration = 7 days;
+
+    uint256 public constant override version = 1;
```
RedFox20, line 29: Version should be the first field

contracts/TempusPool.sol
```diff
@@ -454,6 +460,27 @@ abstract contract TempusPool is ITempusPool {
         return toBackingToken ? backingTokens : yieldTokens;
     }
 
+    /// @dev This updates the underlying pool's interest rate, and also updates the internal
+    ///      tracking of negative interest periods.
+    function updateAndCheckInterestRate() internal returns (uint256 rate) {
+        rate = updateInterestRate();
+
+        if (rate < initialInterestRate) {
+            if (lastNegativeYield == 0) {
```
RedFox20, line 469: Names of the fields are a bit generic, it would be better to use specific names: `negativeYieldStartTime`

contracts/TempusPool.sol
```diff
@@ -195,7 +201,7 @@ abstract contract TempusPool is ITempusPool {
         )
     {
         require(!matured(), "Maturity reached.");
```
RedFox20, line 203: `updateAndCheckInterestRate()` will affect `matured()`, is this the intended ordering of calls? If so, it needs a comment now, otherwise, rate should be done before `matured()` check

contracts/TempusPool.sol
```diff
@@ -289,7 +295,7 @@ abstract contract TempusPool is ITempusPool {
         PrincipalShare(address(principalShare)).burnFrom(from, principalAmount);
         YieldShare(address(yieldShare)).burnFrom(from, yieldAmount);
 
-        uint256 currentRate = updateInterestRate();
+        uint256 currentRate = updateAndCheckInterestRate();
```
RedFox20, line 298: Same issue again, `updateAndCheckInterestRate()` will affect `matured()` state, so is the previous maturity check above correct or not?
Also, this change will automatically finalize the pool at some point due to negative yield

contracts/TempusPool.sol
```diff
@@ -124,7 +130,7 @@ abstract contract TempusPool is ITempusPool {
         returns (uint256 backingTokenAmount);
 
     function matured() public view override returns (bool) {
-        return block.timestamp >= maturityTime;
+        return (block.timestamp >= maturityTime) || (block.timestamp >= exceptionalHaltTime);
```
yuval-weiss, line 133: I think it would be better if we have 2 types of "paused" states - `matured` & `halted`. This makes more sense than merging these 2 conditions into the "matured" state imo.

contracts/TempusPool.sol
axic: Renamed.

contracts/TempusPool.sol
axic: I didn't wanted to change the order because didn't wanted to query the rate after maturity, and figured the extra check for negative yield afterwards would be sufficient -- so a comment is definitely worth it here.

However open for different designs, that `negative yield!` check feels like a bit of an afterthought.

test/TempusPool.Redeem.test.ts
```diff
@@ -141,6 +141,33 @@ describeForEachPool("TempusPool Redeem", (pool:PoolTestFixture) =>
...
+  {
+    await pool.createDefault();
+    let [owner, user] = pool.signers;
+    await pool.setupAccounts(owner, [[user, 200]]);
+
+    await pool.depositYBT(user, 100, /*recipient:*/user);
+    (await pool.userState(user)).expect(100, 100, /*yieldBearing:*/100);
+
+    await pool.setTimeRelativeToPoolStart(24*60*60); // 1 day
+    await pool.setInterestRate(0.9);
```
axic, line 154: Oh, actually need to do an action here to trigger updates.

test/TempusPool.Redeem.test.ts
```diff
@@ -141,6 +141,42 @@ describeForEachPool("TempusPool Redeem", (pool:PoolTestFixture) =>
...
+    await pool.setInterestRate(0.6);
+
+    await pool.redeemToYBT(user, 90, 90);
+    if (pool.yieldPeggedToAsset)
+    {
+        (await pool.userState(user)).expect(0, 0, /*yieldBearing:*/120);
+    }
+    else
+    {
+        (await pool.userState(user)).expect(0, 0, /*yieldBearing:*/200);
```
axic, line 176: This (and other values) needs to be reviewed, as the previous test case ("Should work before maturity with equal shares, with negative yield") arrives at this number too, despite this withdrawing prior.

What was initially confusing is that the accounts are funding with 200 and we only deposit 100.

test/TempusPool.Redeem.test.ts
```diff
@@ -141,6 +141,42 @@ describeForEachPool("TempusPool Redeem", (pool:PoolTestFixture) =>
...
+  it.includeIntegration("Should work before maturity with equal shares, with long period of negative yield", async () =>
+  {
+    await pool.createDefault();
+    let [owner, user] = pool.signers;
+    await pool.setupAccounts(owner, [[user, 200]]);
+
+    await pool.depositYBT(user, 100, /*recipient:*/user);
+    (await pool.userState(user)).expect(100, 100, /*yieldBearing:*/100);
+
+    await pool.setTimeRelativeToPoolStart(24*60*60); // 1 day
```
yuval-weiss, line 153: This should be a relative number (range 0 to 1) to the start of the pool (0.5 to move time to 50% until maturity).
And also - `pool.createDefault` creates a pool with a maturity of 1 hour, so 1 day or 8 days is after maturity

contracts/TempusPool.sol
yuval-weiss: I definitely think the ordering here is problematic... It just means that the pool will never actually be halted on the same transaction where `exceptionalHaltTime` is set because there's no check for that after this variable is set. As I mentioned in another comment, I also think we should separate the `halted` state from the `matured` state (as those are currently merged together).

test/TempusPool.Redeem.test.ts
axic: Ahh, good catch that it is a percentage. Should add a safety check into it.

contracts/TempusPool.sol
```diff
@@ -280,6 +288,7 @@ abstract contract TempusPool is ITempusPool {
         require(IERC20(address(yieldShare)).balanceOf(from) >= yieldAmount, "Insufficient yields.");
 
         if (matured()) {
+            console.log("Matured", (exceptionalHaltTime != type(uint256).max) ? exceptionalHaltTime : 0);
```
yuval-weiss, line 291: remove

contracts/TempusPool.sol
axic: Swapped in #439.

contracts/TempusPool.sol
axic: Swapped in #439.

contracts/TempusPool.sol
```diff
@@ -450,6 +459,31 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable {
         return toBackingToken ? backingTokens : yieldTokens;
     }
 
+    /// @dev This updates the underlying pool's interest rate, and also updates the internal
```
axic, line 462: Based on #428 we may want to name this `updateAndValidateInterestRate` instead.

contracts/TempusPool.sol
```diff
@@ -255,7 +263,7 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable {
 
     function finalize() public override {
         if (matured() && maturityInterestRate == 0) {
-            maturityInterestRate = updateInterestRate();
+            maturityInterestRate = updateAndCheckInterestRate();
```
yuval-weiss, line 266: I think this should just be `updateInterestRate` and `updateAndCheckInterestRate` should be used in `burnShares` (instead of `updateInterestRate`). If `finalize` is called externally it means the pool has matured already anyways...

contracts/TempusPool.sol
yuval-weiss: I think `updateAndValidateInterestRate` is better

axic:
Need to check why Aave aDAI behaves differently:
```
  Aave aDAI <> TempusPool Redeem
    createFixture Aave|aDAI|1|1382400|0.1|0.02|5|5 1174ms
Resetting negative yield period
Entering negative yield period 1638611996
Resetting negative yield period
Entering negative yield period 1639130395
Check for long negative yield period
Exceptionally halting 1639650176
Matured 1639650176 0
Resetting negative yield period
Matured 1639650176 200000000000000000

    1) Should work before maturity with equal shares, with fluctuating yield                        2469ms

  Aave aUSDC <> TempusPool Redeem
    createFixture Aave|aUSDC|1|1382400|0.1|0.02|5|5 503ms
Resetting negative yield period
Entering negative yield period 1638776156
Resetting negative yield period
Entering negative yield period 1639787036
Check for long negative yield period
Resetting negative yield period
    ✓ Should work before maturity with equal shares, with fluctuating yield                         796ms

  Compound cDAI <> TempusPool Redeem
    createFixture Compound|cDAI|1|1382400|0.1|0.02|5|5 502ms
Resetting negative yield period
Entering negative yield period 1638932110
Resetting negative yield period
Entering negative yield period 1640410846
Check for long negative yield period
Resetting negative yield period
    ✓ Should work before maturity with equal shares, with fluctuating yield                         789ms

  Compound cUSDC <> TempusPool Redeem
    createFixture Compound|cUSDC|1|1382400|0.1|0.02|5|5 497ms
Resetting negative yield period
Entering negative yield period 1639080266
Resetting negative yield period
Entering negative yield period 1641003465
Check for long negative yield period
Resetting negative yield period
    ✓ Should work before maturity with equal shares, with fluctuating yield                         764ms

  Lido stETH <> TempusPool Redeem
    createFixture Lido|stETH|1|1382400|0.1|0.02|5|5 425ms
Resetting negative yield period
Entering negative yield period 1639221012
Check for long negative yield period
Check for long negative yield period
Check for long negative yield period
Resetting negative yield period
    ✓ Should work before maturity with equal shares, with fluctuating yield                         731ms

  Yearn yvDAI <> TempusPool Redeem
    createFixture Yearn|yvDAI|1|1382400|0.1|0.02|5|5 475ms
Resetting negative yield period
Entering negative yield period 1639354722
Resetting negative yield period
Entering negative yield period 1642101291
Check for long negative yield period
Resetting negative yield period
    ✓ Should work before maturity with equal shares, with fluctuating yield                         750ms

  Yearn yvUSDC <> TempusPool Redeem
    createFixture Yearn|yvUSDC|1|1382400|0.1|0.02|5|5 453ms
Resetting negative yield period
Entering negative yield period 1639481747
Resetting negative yield period
Entering negative yield period 1642609387
Check for long negative yield period
Resetting negative yield period
    ✓ Should work before maturity with equal shares, with fluctuating yield                         704ms
```

axic:
After rebasing this on #445, which means there is no period exceeding 7 days of negative interest (before that by mistake the period was 7+ days), the results are the follow (lido fails, because it matures):
```
  Aave aDAI <> TempusPool Redeem
Initial interest rate 1000000000000000000
    createFixture Aave|aDAI|1|1382400|0.1|0.02|5|5 1129ms
Resetting negative yield period
Entering negative yield period 1638582680 900000000000000000
Resetting negative yield period
Entering negative yield period 1639187480 600000000000000000
Check for long negative yield period 1639792280 1639619480 200000000000000000
Resetting negative yield period
    ✓ Should work before maturity with equal shares, with fluctuating yield                         2232ms

  Aave aUSDC <> TempusPool Redeem
Initial interest rate 1000000000000000000
    createFixture Aave|aUSDC|1|1382400|0.1|0.02|5|5 500ms
Resetting negative yield period
Entering negative yield period 1638582680 900000000000000000
Resetting negative yield period
Entering negative yield period 1639187480 600000000000000000
Check for long negative yield period 1639792280 1639619480 200000000000000000
Resetting negative yield period
    ✓ Should work before maturity with equal shares, with fluctuating yield                         777ms

  Compound cDAI <> TempusPool Redeem
Initial interest rate 10000000000000000000000000000
    createFixture Compound|cDAI|1|1382400|0.1|0.02|5|5 491ms
Resetting negative yield period
Entering negative yield period 1638582682 9000000000000000000000000000
Resetting negative yield period
Entering negative yield period 1639187482 6000000000000000000000000000
Check for long negative yield period 1639792282 1639619482 2000000000000000000000000000
Resetting negative yield period
    ✓ Should work before maturity with equal shares, with fluctuating yield                         766ms

  Compound cUSDC <> TempusPool Redeem
Initial interest rate 10000000000000000
    createFixture Compound|cUSDC|1|1382400|0.1|0.02|5|5 485ms
Resetting negative yield period
Entering negative yield period 1638582683 9000000000000000
Resetting negative yield period
Entering negative yield period 1639187483 6000000000000000
Check for long negative yield period 1639792283 1639619483 2000000000000000
Resetting negative yield period
    ✓ Should work before maturity with equal shares, with fluctuating yield                         759ms

  Lido stETH <> TempusPool Redeem
Initial interest rate 1000000000000000000
    createFixture Lido|stETH|1|1382400|0.1|0.02|5|5 423ms
Resetting negative yield period
Entering negative yield period 1638582681 900000000000000000
Check for long negative yield period 1639187481 1638928281 999999999999999999
Check for long negative yield period 1639187481 1639187481 599999999999999999
Exceptionally halting 1639187481 599999999999999999
Matured 1639187481 0
Check for long negative yield period 1639187481 1639619481 199999999999999999
Exceptionally halting 1639619481 199999999999999999
Matured 1639619481 599999999999999999
Resetting negative yield period
Matured 1639619481 599999999999999999

    1) Should work before maturity with equal shares, with fluctuating yield                        718ms

  Yearn yvDAI <> TempusPool Redeem
Initial interest rate 1000000000000000000
    createFixture Yearn|yvDAI|1|1382400|0.1|0.02|5|5 481ms
Resetting negative yield period
Entering negative yield period 1638582683 900000000000000000
Resetting negative yield period
Entering negative yield period 1639187483 600000000000000000
Check for long negative yield period 1639792283 1639619483 200000000000000000
Resetting negative yield period
    ✓ Should work before maturity with equal shares, with fluctuating yield                         770ms

  Yearn yvUSDC <> TempusPool Redeem
Initial interest rate 1000000
    createFixture Yearn|yvUSDC|1|1382400|0.1|0.02|5|5 580ms
Resetting negative yield period
Entering negative yield period 1638582684 900000
Resetting negative yield period
Entering negative yield period 1639187484 600000
Check for long negative yield period 1639792284 1639619484 200000
Resetting negative yield period
    ✓ Should work before maturity with equal shares, with fluctuating yield                         891ms
```

axic:
Ok I think the reason is that Lido's rate setting is not precise enough and instead of `1` it ends ups at `0.999999999999999999` which is still considered negative.

I changed the test cases to do `1.1` as the non-negative rate and now it seems to be correct.

contracts/TempusPool.sol
```diff
@@ -447,6 +463,42 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable {
         return toBackingToken ? backingTokens : yieldTokens;
     }
 
+    /// @dev This updates the underlying pool's interest rate, and also updates the internal
+    ///      tracking of negative interest periods.
+    function updateAndValidateInterestRate() internal returns (uint256 rate) {
+        rate = updateInterestRate();
+
+        // Short circuit. No need for the below after maturity.
+        if (matured()) {
```
axic, line 472: This is a new feature, to avoid doing any calculation once matured.

contracts/TempusPool.sol
```diff
@@ -28,11 +28,14 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable {
 
     uint256 public constant override version = 1;
 
+    uint256 public constant override maximumNegativeYieldDuration = 7 days;
```
axic, line 31: Should make this configurable, in a new PR.

contracts/ITempusPool.sol
```diff
@@ -77,7 +77,16 @@ interface ITempusPool is ITempusFees {
     /// @return Maturity time of the pool.
     function maturityTime() external view returns (uint256);
 
+    /// @return Time of exceptional halting of the pool.
+    /// In case the pool is still in operation, this must return type(uint256).max.
+    function exceptionalHaltTime() external view returns (uint256);
+
+    /// @return The maximum allowed time (in seconds) to pass with negative yield.
+    function maximumNegativeYieldDuration() external view returns (uint256);
```
axic, line 85: Should add basic tests for these two.

test/utils/TempusPool.ts
```diff
@@ -398,6 +398,27 @@ export class TempusPool extends ContractBase {
...
   }
 
+  /**
+   * @returns The exceptional halt time of the pool
+   * @note This returns null in case it is not set (i.e. has the special value of `type(uin256).max`)
+   */
+  async exceptionalHaltTime(): Promise<NumberOrString | null> {
+    let exceptionalHaltTime:BigNumber = await this.contract.exceptionalHaltTime();
+//    if (exceptionalHaltTime.eq(BigNumber.from(MAX_UINT256))) {
+    if (exceptionalHaltTime.toHexString() === MAX_UINT256) {
```
axic, line 408: Here are two options to accomplish this. Also could just use a local string constant instead of important this one.

contracts/TempusPool.sol
axic: See #449.

RedFox20 (dismissed):

RedFox20 (approved):

PR #427, opened 2021-11-11
test: Negative yield redeem before maturity

axic:


test/TempusPool.Redeem.test.ts
```diff
@@ -119,6 +119,28 @@ describeForEachPool("TempusPool Redeem", (pool:PoolTestFixture) =>
     }
   });
 
+  it.includeIntegration("Should work before maturity with equal shares, with negative yield", async () =>
```
axic, line 122: This case was missing.

mijovic (approved):

PR #428, opened 2021-11-14
Interest rate manipulation protection

yuval-weiss:


contracts/TempusPool.sol
```diff
@@ -343,6 +349,28 @@ abstract contract TempusPool is ITempusPool {
...
+        uint256 latestRate = updateInterestRate();
+        /// Sanity check the interest rate doesn't deviate too much from the EMA normalized rate
+        /// (due to some kind of interest rate manipulation attack)
+        // require(latestRate /emaValue < maxDiff)
+        uint256 forecastedCurrentInterestRate = forecastInterestAtTimestamp(block.timestamp);
+        uint256 forecastedRateDiff = latestRate > forecastedCurrentInterestRate ? (latestRate - forecastedCurrentInterestRate) : (forecastedCurrentInterestRate - latestRate); 
+        
+        uint256 forecastedRateDeviation = forecastedRateDiff.divfV(forecastedCurrentInterestRate, MAX_FORECASTED_RATE_DEVIATION_TOLERANCE_DENOMINATOR);
+        
+        /// TODO: IMPORTANT - maybe add logic that remove ignores this condition in case the forecasted rate
```
yuval-weiss, line 362: This is really important and needs further thought & discussion. We must ensure there's no way that TempusPools will become unusable if there's some issue with this rate manipulation protection mechanism... It exists to protect against flash loan attacks that are temporary by nature, so if this require statement reverts several times in a row we should just disable this protection imo...

contracts/SmoothedRateOracle.sol
```diff
@@ -0,0 +1,76 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.6;
+
+import "./math/Fixed256xVar.sol";
+
+/// Implements https://en.wikipedia.org/wiki/Exponential_smoothing#Double_exponential_smoothing
+abstract contract SmoothedRateOracle {
```
yuval-weiss, line 7: This uses Double Exponential Smoothing rather than Simple Exponential Smoothing (as we discussed @RedFox20). I chose Double Exponential Smoothing because the interest rate is always increasing by nature so Simple Exponential Smoothing isn't really suitable... Double Exponential Smoothing on the other hand takes trends into account.

contracts/SmoothedRateOracle.sol
```diff
@@ -0,0 +1,76 @@
...
+    /// TODO: IMPORTANT use struct and try to fit
+    ///     latestSmoothedRate, latestTrend & latestSmoothedRateSampleTimestamp in 1-2 storage slots instead of 3
+    /// Note - this will require checking if smaller variables than uint256 can be used (prolly yeah)
+    uint256 internal latestSmoothedRate;
+    uint256 internal latestTrend;
+    uint256 internal latestSmoothedRateSampleTimestamp;
+
+    /// TODO: IMPORTANT these need to be optimized.
+    // Hyperparams
+    uint256 private constant LEVEL_SMOOTHING_MULTIPLIER = 10;
```
yuval-weiss, line 19: I chose these randomly for now

contracts/SmoothedRateOracle.sol
```diff
@@ -0,0 +1,76 @@
...
+        }
+    }
+
+    /// F[t+m] = S[t] + m * B(t)
+    function forecastInterestAtTimestamp(uint256 timestamp) internal view returns (uint256) {
+        return latestSmoothedRate +
+            (timestamp - latestSmoothedRateSampleTimestamp).mulfV(latestTrend, EMA_SAMPLE_PERIOD);
+    }
+
+    function setSmoothedRate(uint256 smoothedRate, uint256 trend) private {
```
yuval-weiss, line 71: We can optimize that and fit these 3 params into 1-2 storage slots probably. This would save us a lot of gas whenever we do an update.

axic:
@yuval-weiss what is the plan with this?

mijovic:
I don't think this is very important, at least for current integrations. We can revisit if it is needed in future

PR #431, opened 2021-11-19
Introduce Staking

axic:


contracts/token/staking/StakingToken.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.10;
+
+import "@openzeppelin/contracts/token/ERC20/extensions/ERC20Votes.sol";
+import "../../utils/Ownable.sol";
+
+/// Token representing a staking position.
+contract StakingToken is Ownable, ERC20Votes {
```
RedFox20, line 8: Name should probably be `StakedTempusToken.sol` for consistency

PR #432, opened 2021-11-21
sanity checks

yuval-weiss:


yuval-weiss:
I think we should add a setting, that can be invoked by the contract owner, which disables/enables these sanity checks. I think it's good to have these checks, but I'm kinda worried that we have a scenario where for some unknown reason these sanity checks fail, which causes funds to be forever locked in the contract 🤮.

mijovic:
> I think we should add a setting, that can be invoked by the contract owner, which disables/enables these sanity checks. I think it's good to have these checks, but I'm kinda worried that we have a scenario where for some unknown reason these sanity checks fail, which causes funds to be forever locked in the contract 🤮.

I think there is a better way to do this, without introducing new owner methods. Can we just do these checks on deposit? It is completely fine to leave deposits locked, and redeems should work always.

contracts/TempusPool.sol
```diff
@@ -343,6 +347,20 @@ abstract contract TempusPool is ITempusPool {
...
+    function sanityChecks() private view {
+        /// TODO: maybe do some validations using pricePerShareStored before calling updateInterestRate ?
+        uint256 totalPrincipals = IERC20Metadata(address(principalShare)).totalSupply();
+        uint256 totalYields = IERC20Metadata(address(yieldShare)).totalSupply();
+
+        uint256 totalYieldBearingBalance = IERC20(yieldBearingToken).balanceOf(address(this));
+        uint256 totalBackingTokensInControl = numAssetsPerYieldToken(totalYieldBearingBalance, currentInterestRate());
+
+        uint256 totalYieldsValue = pricePerYieldShareStored().mulfV(totalYields, backingTokenONE);
+        uint256 totalPrincipalsValue = pricePerPrincipalShareStored().mulfV(totalPrincipals, backingTokenONE);
```
mijovic, line 359: Maybe using `getRedemptionAmounts` instead of relying on `pricePerShare`?

yuval-weiss:
> 

We could, I considered that as well, but it would be nice to have this check on redemption as well (to validate we didn't burn too little shares)

mijovic:
Closing, we can revisit this later. I think it is even not neccesary

PR #434, opened 2021-11-23
test: Add assertion that percentage inputs are correct

axic:
Unfortunately there's no range type in Typescript yet: https://stackoverflow.com/questions/39494689/is-it-possible-to-restrict-number-to-a-certain-range


test/pool-utils/PoolTestFixture.ts
```diff
@@ -271,6 +272,7 @@ export abstract class PoolTestFixture {
    * Fast forwards time to certain point in time
    */
   async setTimeRelativeToPoolStart(percentDuration: number): Promise<void> {
+    assert.ok(percentDuration >= 0.0 && percentDuration <= 1.0);
```
axic, line 275: This isn't the nicest, but at least helps us see if it was misused anywhere. Feel free not to merge t

yuval-weiss (approved):

mijovic (approved):

PR #438, opened 2021-11-26
pool: Make ITempusFees/TempusPool properly ownable

axic:
This became more complicated than expected.

### DIFF
--- a/contracts/ITempusPool.sol
+++ b/contracts/ITempusPool.sol
@@ -3,9 +3,11 @@ pragma solidity >=0.7.6 <0.9.0;
 pragma abicoder v2;
 
 import "./token/IPoolShare.sol";
+import "./utils/IOwnable.sol";
 
-interface ITempusFees {
-    // The fees are in terms of yield bearing token (YBT).
+/// Setting and transferring of fees are restricted to the owner.
+interface ITempusFees is IOwnable {
+    /// The fees are in terms of yield bearing token (YBT).
     struct FeesConfig {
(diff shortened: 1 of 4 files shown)


contracts/ITempusPool.sol
```diff
@@ -3,8 +3,9 @@ pragma solidity >=0.7.6 <0.9.0;
 pragma abicoder v2;
 
 import "./token/IPoolShare.sol";
+import "./utils/IOwnable.sol";
 
-interface ITempusFees {
+interface ITempusFees is IOwnable {
```
axic, line 8: `IOwnable` could be on `ITempusPool`, but I felt that logically it is connected to fees. Should/could leave a comment about this.

mijovic: Hm, maybe it is better to add comment, it is clear now, but in some time it might happen we forget about it

contracts/TempusPool.sol
```diff
@@ -139,7 +133,7 @@ abstract contract TempusPool is ITempusPool {
     }
 
     function transferFees(address authorizer, address recipient) external override onlyController {
-        require(authorizer == owner, "Only owner can transfer fees.");
+        require(authorizer == owner(), "Only owner can transfer fees.");
```
mijovic, line 136: I guess in next pr this is removed and we use modifier instead

axic: Yes, the next PR is the previous is number: #437.

mijovic (approved):

PR #441, opened 2021-11-28
Rari Stable Pools integration

yuval-weiss:
TODOS:
* Handle `initialInterestRate` rebasing when the Rari Pool Withdrawal Fee changes (cache `_rariWithdrawalFee` and check for changes).
* Unit tests framework integration (implementing `RariTestPool` and other related TS classes & logic).
* Documentation.


integration-tests/TempusPool.rari.test.ts
```diff
@@ -0,0 +1,200 @@
...
+        }
+      };
+    })
+  }
+
+  return fixtures[rariFee.toString()]()
+}
+
+describe('TempusPool <> Rari', function () {
+  /// TODO: IMPORTANT we need to handle (and test) the case where the Rari fee changes
```
yuval-weiss, line 74: This is important sers.

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,165 @@
...
+            yieldBearingToken,
+            _backingTokenRariPoolIndex
+        );
+
+        return currentInterestRate();
+    }
+
+    /// @return Stored Interest Rate with the same precision as the BackingToken
+    function currentInterestRate() public view override returns (uint256) {
+        return _lastCalculatedInterestRate;
```
yuval-weiss, line 113: A note about this - there's an issue with this and that is that if there are no deposits/redeems on the RariTempusPool for some time, that means that `currentInterestRate` will return a stale interest rate. This could be a problem because dApps will use a stale exchange rate between Yields & Principals (using `TempusAMM.getExpectedReturnGivenIn`, which relies on `currentInterestRate`) which might cause swaps to fail to due high slippage.

To solve this I think we can either make `TempusPool. updateInterestRate` public, or just expose it on the RariTempusPool (for example by making `calculateInterestRate` public).

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,165 @@
...
+        returns (uint256 backingTokenAmount)
+    {
+        uint256 rftTotalSupply = IERC20(yieldBearingToken).totalSupply();
+        uint256 withdrawalAmountUsd = (yieldBearingTokensAmount * rariFundManager.getFundBalance()) / rftTotalSupply;
+
+        uint256 backingTokenToUsdRate = rariFundManager.rariFundPriceConsumer().getCurrencyPricesInUsd()[
+            _backingTokenRariPoolIndex
+        ];
+
+        uint256 withdrawalAmountInBackingToken = withdrawalAmountUsd.mulfV(backingTokenONE, backingTokenToUsdRate);
```
RedFox20, line 84: Huh, this is interesting, I would have expected `withdrawalAmountUsd.mulfV(backingTokenToUsdRate, backingTokenONE)` instead

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,165 @@
...
+        if (withdrawalAmountInBackingToken.mulfV(backingTokenToUsdRate, backingTokenONE) > withdrawalAmountUsd) {
```
RedFox20, line 85: Like here 😕 Wtf is going on here 😄

contracts/pools/RariTempusPool.sol
RedFox20: The first action of AMM/TempusPool should always be to call updateInterestRate() and then later use currentInterestRate() when needing access to it. So you should check if AMM and TempusPool handle this properly

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,165 @@
...
+
+        uint256 preFeeRate;
+        if (rftTotalSupply > 0 && fundBalanceUsd > 0) {
+            preFeeRate = backingTokenToUsdRate.mulfV(fundBalanceUsd, rftTotalSupply);
+        } else {
+            preFeeRate = backingTokenToUsdRate;
+        }
+
+        /// Apply fee
+        uint256 postFeeRate = preFeeRate.mulfV(1e18 - fundManager.getWithdrawalFeeRate(), 1e18);
```
RedFox20, line 146: Is getWithdrawalFeeRate() always 1e18 ? And interest rate itself always 1e18 ? Just making sure

integration-tests/TempusPool.compound.test.ts
```diff
@@ -136,7 +136,7 @@ describe('TempusPool <> Compound <> DAI', function () {
     
     const error = new Decimal(1).sub(new Decimal(fromWei(totalInterestSigner2).toString())
       .div(fromWei(totalInterestSigner1).toString())).abs()
```
RedFox20, line 138: `abs();`

RedFox20:
This needs: 
- [ ] support in TempusPool.ts
- [ ] support in MultiPoolTestSuite.ts
- [ ] support in Config.ts

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,165 @@
...
+    /// @dev The rate precision always matches the BackingToken's precision
+    function interestRateToSharePrice(uint interestRate) internal pure override returns (uint) {
+        return interestRate;
+    }
+
+    function calculateInterestRate(
+        IRariFundManager fundManager,
+        address ybToken,
+        uint256 currencyIndex
+    ) private returns (uint256) {
```
axic, line 133: `view` or `pure` or is this writing state?

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        /// Apply fee
+        uint256 postFeeRate = preFeeRate.mulfV(1e18 - fundManager.getWithdrawalFeeRate(), 1e18);
+
+        return postFeeRate;
+    }
+
+    function getTokenRariPoolIndex(IRariFundManager fundManager, address bToken) private view returns (uint256) {
+        string[] memory acceptedSymbols = fundManager.getAcceptedCurrencies();
```
axic, line 166: Since all of this is done based on symbols only, it would be nice have some reassurance at the end that the address matches that of the token. Even if it is not possible to iterate addresses, is it possible to validate them?

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,179 @@
...
+        string memory backingTokenSymbol = IERC20Metadata(bToken).symbol();
+
+        uint256 backingTokenIndex = type(uint256).max;
+        for (uint256 i = 0; i < acceptedSymbols.length; i++) {
+            if (keccak256(abi.encodePacked(backingTokenSymbol)) == keccak256(abi.encodePacked(acceptedSymbols[i]))) {
```
axic, line 171: I think this should work too as the actual encoding style is not important as long as the types are the same:
```suggestion
            if (keccak256(bytes(backingTokenSymbol)) == keccak256(bytes(acceptedSymbols[i]))) {
```

This should save gas due to lower memory usage.

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,179 @@
...
+            estYield,
+            principalsData,
+            yieldsData,
+            maxFeeSetup
+        )
+    {
+        rariFundManager = fundManager;
+
+        uint256 backingTokenIndex = getTokenRariPoolIndex(fundManager, backingToken);
+        require(backingTokenIndex != type(uint256).max, "backing token is not accepted by the rari pool");
```
axic, line 56: Could also put this require statement into the helper, or use two variables (a second boolean for failure as that is a more common pattern), but I think this approach is good too.

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    function depositToUnderlying(uint256 amount) internal override returns (uint256) {
+        require(msg.value == 0, "ETH deposits not supported");
+
+        // Deposit to Rari Pool
+        IERC20(backingToken).safeIncreaseAllowance(address(rariFundManager), amount);
+
+        uint256 preDepositBalance = IERC20(yieldBearingToken).balanceOf(address(this));
+        rariFundManager.deposit(IERC20Metadata(backingToken).symbol(), amount);
```
axic, line 74: I mean, really? This symbol stuff is crazy.

Also should cache the symbol as an immutable if possible.

axic: Doesn't this return anything? What if the symbol is not found, will it revert?

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,179 @@
...
+        uint256 backingTokenToUsdRate = rariFundManager.rariFundPriceConsumer().getCurrencyPricesInUsd()[
+            _backingTokenRariPoolIndex
+        ];
+
+        uint256 withdrawalAmountInBackingToken = withdrawalAmountUsd.mulfV(backingTokenONE, backingTokenToUsdRate);
+        if (withdrawalAmountInBackingToken.mulfV(backingTokenToUsdRate, backingTokenONE) > withdrawalAmountUsd) {
+            withdrawalAmountInBackingToken -= 1; /// Round down if there were any rounding errors
+        }
+
+        uint256 amountWithdrawn = rariFundManager.withdraw(
```
axic, line 97: In deposit you manually check balances -- would it make sense double checking it here too?

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,179 @@
...
+            IERC20Metadata(backingToken).symbol(),
+            withdrawalAmountInBackingToken
+        );
+        if (recipient != address(this)) {
```
axic, line 101: Hmm, this check is inconsistent with the other pool implementations (they ignore this). However this makes me wonder if we should add a require statement that `recipient != address(this)`?

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,179 @@
...
+        /// return interestRate; /// TODO: IMPORTANT
+    }
+
+    function calculateInterestRate(
+        IRariFundManager fundManager,
+        address ybToken,
+        uint256 currencyIndex
+    ) private returns (uint256) {
+        uint256 backingTokenToUsdRate = fundManager.rariFundPriceConsumer().getCurrencyPricesInUsd()[currencyIndex];
+        uint256 rftTotalSupply = IERC20(ybToken).totalSupply();
```
axic, line 148: What is `rft`?

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,179 @@
...
+        uint256 fundBalanceUsd = rftTotalSupply > 0 ? fundManager.getFundBalance() : 0; // Only set if used
```
axic, line 149: Is the ternary an optimisation or the fund returns invalid value on `totalSupply == 0`?

contracts/pools/RariTempusPool.sol
yuval-weiss: Hmm unfortunately no - https://github.com/Rari-Capital/rari-stable-pool-contracts/blob/386aa8811e7f12c2908066ae17af923758503739/contracts/RariFundManager.sol#L89.
The addresses mapping is private. If we were to accidentally initialize a RariTempusPool with a fake backingToken that has the same symbol as one of Rari's supported currencies, it would just fail on deposits.

contracts/pools/RariTempusPool.sol
yuval-weiss: It's potentially writing state. Specifically because of the `fundManager.getFundBalance` call.

contracts/pools/RariTempusPool.sol
yuval-weiss: Yeah, both are always 18.

contracts/pools/RariTempusPool.sol
yuval-weiss: It's an optimization (ref - https://github.com/Rari-Capital/rari-stable-pool-contracts/blob/386aa8811e7f12c2908066ae17af923758503739/contracts/RariFundManager.sol#L591). There's no need to query the fund balance if the YBT supply is 0.

contracts/pools/RariTempusPool.sol
yuval-weiss: Rari's yield bearing tokens

contracts/pools/RariTempusPool.sol
RedFox20: cool! 😄

contracts/pools/RariTempusPool.sol
yuval-weiss: Ah you're right. I think we definitely should add a require statement for that. But for now I will just remove this condition from the RariTempusPool

contracts/pools/RariTempusPool.sol
yuval-weiss: The reason I added explicit pre and post deposit balance checks is that Rari's `deposit` function doesn't return the minted YBT amount...

yuval-weiss: But honestly, I think all of the TempusPool implementations should manually check balances before and after rather than trusting when 3rd party protocols returned... What do you think?

contracts/pools/RariTempusPool.sol
yuval-weiss: I agree. Will add an immutable variable for the symbol.

yuval-weiss: Yeah, in case the symbol is not found, deposit() will revert (https://github.com/Rari-Capital/rari-stable-pool-contracts/blob/386aa8811e7f12c2908066ae17af923758503739/contracts/RariFundManager.sol#L576)

yuval-weiss: hmmm actually when I think of it, why add another immutable variable for caching the symbol? Right now we're just calling `backingToken.symbol()`, and `backingToken` is an immutable anyways.

contracts/pools/RariTempusPool.sol
yuval-weiss: I agree, better to move the require statement inside the function.

contracts/pools/RariTempusPool.sol
axic: `.symbol()` is an actual function call.

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    function getTokenRariPoolIndex(IRariFundManager fundManager, address bToken) private view returns (uint256) {
+        string[] memory acceptedSymbols = fundManager.getAcceptedCurrencies();
+        string memory backingTokenSymbol = IERC20Metadata(bToken).symbol();
+    
+        uint256 backingTokenIndex = type(uint256).max;
+        for (uint256 i = 0; i < acceptedSymbols.length; i++) {
+            if (keccak256(bytes(backingTokenSymbol)) == keccak256(bytes(acceptedSymbols[i]))) {
+                backingTokenIndex = i;
```
axic, line 165: Could change this to be `return i;`, and then you don't need the `backingTokenIndex` variable and the trailing statement would be `revert("backing token is not accepted...");`

axic: It is a tiny gas saving, but more importantly it is susceptible to the problem "what if rare returns max as index?" :)

axic: Instead of `return i;` could also name the return variable `backingTokenIndex` and do
```
backingTokenIndex = i;
return;
```

yuval-weiss: true

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,174 @@
...
+    function currentInterestRate() public view override returns (uint256) {
+        return _lastCalculatedInterestRate;
+    }
+
+    function numAssetsPerYieldToken(uint yieldTokens, uint rate) public view override returns (uint) {
+        return yieldTokens.mulfV(rate, exchangeRateONE) / exchangeRateToBackingPrecision;
+    }
+
+    function numYieldTokensPerAsset(uint backingTokens, uint rate) public view override returns (uint) {
+        return backingTokens.divfV(rate, exchangeRateONE) * exchangeRateToBackingPrecision;
```
axic, line 128: I wonder about precision here as we first divide then multiply. Would it fit 256-bits to first multiply?

yuval-weiss: We don't divide first here... divfV here is equivalent to backingTokens * exchangeRateONE / rate

test/utils/PoolShare.ts
```diff
@@ -24,9 +24,17 @@ export class PoolShare extends ERC20OwnerMintable {
...
    */
   async getPricePerFullShareStored(): Promise<NumberOrString> {
     return this.fromBigNum(await this.contract.getPricePerFullShareStored());
   }
+
+  /**
+   * @returns Price per share as described in PoolShare.sol
+   * @notice - This is a state-changing method
+   */
+   async getPricePerFullShare(): Promise<Transaction> {
```
axic, line 37: Can this actually return a value?

axic: If this is only about interest rate update, I'd call it differently :)

yuval-weiss: PoolShare.ts is just a wrapper, so it just reflects the same name of the contract function. I wouldn't name it differently but an alternative for this would be to just remove this function completely from PoolShare.ts and then in `RariTestPool` we can replace `await this.yields.getPricePerFullShare();` with `await this.yields.contract.getPricePerFullShare();`, which kinda makes sense as well because it's just a temporary hack

axic: Moving it closer I think is a good idea.

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,172 @@
...
+    using SafeERC20 for IERC20;
+    using UntrustedERC20 for IERC20;
+    using Fixed256xVar for uint256;
+
+    bytes32 public immutable override protocolName = "Rari";
+    IRariFundManager internal immutable rariFundManager;
+
+    uint256 private immutable exchangeRateToBackingPrecision;
+    uint256 private immutable _backingTokenRariPoolIndex;
+    uint256 private _lastCalculatedInterestRate;
```
axic, line 23: Why are these underscored? The other pool implementations do not use underscores here.

integration-tests/TempusPool.yearn.test.ts
```diff
@@ -98,8 +98,8 @@ describe('TempusPool <> Yearn', function () {
     const totalInterestSigner2 = toWei(btBalancePostSigner2).sub(toWei(btBalancePreSigner2));
     
     const error = new Decimal(1).sub(new Decimal(fromWei(totalInterestSigner2).toString())
-      .div(fromWei(totalInterestSigner1).toString())).abs()
-    
+      .div(fromWei(totalInterestSigner1).toString())).abs();
```
axic, line 101: Freaking javascript/typescript. Having the linter on tests would be nice :trollface:

yuval-weiss: We can add ESLint

contracts/pools/RariTempusPool.sol
yuval-weiss: True. Fixed.

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,172 @@
...
+import "../protocols/rari/IRariFundManager.sol";
+import "../utils/UntrustedERC20.sol";
+import "../math/Fixed256xVar.sol";
+
+contract RariTempusPool is TempusPool {
+    using SafeERC20 for IERC20;
+    using UntrustedERC20 for IERC20;
+    using Fixed256xVar for uint256;
+
+    bytes32 public immutable override protocolName = "Rari";
```
axic, line 18:
```suggestion
    bytes32 public constant override protocolName = "Rari";
```

The immutable just wastes extra bytecode here.

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,172 @@
...
+            yieldsData,
+            maxFeeSetup
+        )
+    {
+        require(
+            IERC20Metadata(yieldBearingToken).decimals() == 18,
+            "only 18 decimal Rari Yield Bearing Tokens are supported"
+        );
+
+        uint256 backingTokenIndex = getTokenRariPoolIndex(fundManager, backingToken);
```
axic, line 57: Oh, this is actually executed twice in the constructor. Too bad.

The only way to avoid it by moving the `TempusPool(...)` thing down here as `super(...)`, which may not be better to read.

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,172 @@
...
+            1e18,
+            estYield,
+            principalsData,
+            yieldsData,
+            maxFeeSetup
+        )
+    {
+        require(
+            IERC20Metadata(yieldBearingToken).decimals() == 18,
+            "only 18 decimal Rari Yield Bearing Tokens are supported"
```
axic, line 54: Why is this the case? Can you leave a comment?

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,172 @@
...
+        uint256 rftTotalSupply = IERC20(ybToken).totalSupply();
+        uint256 fundBalanceUsd = rftTotalSupply > 0 ? fundManager.getFundBalance() : 0; // Only set if used
+
+        uint256 preFeeRate;
+        if (rftTotalSupply > 0 && fundBalanceUsd > 0) {
+            preFeeRate = backingTokenToUsdRate.mulfV(fundBalanceUsd, rftTotalSupply);
+        } else {
+            preFeeRate = backingTokenToUsdRate;
+        }
+
```
axic, line 153: Rari has a require statement here that `preFeeRate != 0`. Perhaps doing the same out of caution?

contracts/mocks/rari/RariFundTokenMock.sol
```diff
@@ -0,0 +1,22 @@
...
+
+// SPDX-License-Identifier: GPL-3.0
+
+pragma solidity 0.8.10;
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+
+contract RariFundTokenMock is ERC20 {
```
axic, line 11: Can you use the `ERC20OwnerMintable` instead of this (as in other mocks/tests)?

contracts/pools/RariTempusPool.sol
yuval-weiss: Hmmm yeah that's a good point. Both options are not ideal but I'd say let's leave it as is...

contracts/mocks/rari/RariFundManagerMock.sol
```diff
@@ -0,0 +1,161 @@
...
+        string memory backingTokenSymbol = _asset.symbol();
+        string[] memory acceptedSymbols = getAcceptedCurrencies();
+        for (uint256 i = 0; i < acceptedSymbols.length; i++) {
+            if (keccak256(bytes(backingTokenSymbol)) == keccak256(bytes(acceptedSymbols[i]))) {
+                backingTokenIndex = i;
+                break;
+            }
+        }
+
+        require(backingTokenIndex != type(uint256).max, "backing token is not accepted by the rari pool mock");
```
RedFox20, line 45: `not accepted by Rari pool`

contracts/pools/RariTempusPool.sol
RedFox20: Yeah, this should be an immutable field

yuval-weiss: It's not possible to have immutable strings

contracts/mocks/rari/RariFundTokenMock.sol
yuval-weiss: Yup, done.

contracts/pools/RariTempusPool.sol
yuval-weiss: Added to `updateInterestRate`.

contracts/pools/RariTempusPool.sol
yuval-weiss: Added.

contracts/pools/RariTempusPool.sol
yuval-weiss: Done for both Rari and Yearn.

contracts/pools/RariTempusPool.sol
yuval-weiss: Done ser

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,165 @@
...
+            fundManager.rariFundToken(),
+            backingToken,
+            controller,
+            maturity,
+            calculateInterestRate(
+                fundManager,
+                fundManager.rariFundToken(),
+                getTokenRariPoolIndex(fundManager, backingToken)
+            ),
+            10**(IERC20Metadata(backingToken).decimals()),
```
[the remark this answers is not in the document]
yuval-weiss: Done

test/utils/PoolShare.ts
axic: @yuval-weiss can you also do this in case you haven't?

contracts/pools/RariTempusPool.sol
axic: I did re-read the Rari side of this and thought we can make this entire block look nicer, but not really :)

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,177 @@
+// SPDX-License-Identifier: UNLICENSED
```
axic, line 1: Perhaps use the new one in this PR?

contracts/pools/RariTempusPool.sol
axic: I'd suggest to give it a try with storing it as `bytesNN`, if that doesn't work or is messy, I'll do a gas comparison between storage vs. call, though my impression is long term the call is cheaper.

contracts/pools/RariTempusPool.sol
axic: Not sure about that, `calculateInterestRate` is also used in the constructor, so better to have the check here.

contracts/protocols/rari/IRariFundPriceConsumer.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity 0.8.10;
+
+/***
```
axic, line 4: These are wrong natspec style comments, it should be either 
```
/// @notice based on https://github.com/Rari-Capital/rari-stable-pool-contracts/blob/386aa8811e7f12c2908066ae17af923758503739/contracts/RariFundPriceConsumer.sol
```

or

```
/**
  * @notice based on https://github.com/Rari-Capital/rari-stable-pool-contracts/blob/386aa8811e7f12c2908066ae17af923758503739/contracts/RariFundPriceConsumer.sol
  */
```

contracts/protocols/rari/IRariFundPriceConsumer.sol
```diff
@@ -0,0 +1,9 @@
...
+    @notice based on https://github.com/Rari-Capital/rari-stable-pool-contracts/blob/386aa8811e7f12c2908066ae17af923758503739/contracts/RariFundPriceConsumer.sol
+ */
+interface IRariFundPriceConsumer {
+    function getCurrencyPricesInUsd() external view returns (uint256[] memory);
```
axic, line 8: Perhaps add a `/// @return ...` comment, and/or a variable name for the return type.

contracts/protocols/rari/IRariFundManager.sol
```diff
@@ -0,0 +1,35 @@
...
+    /// @return The Rari Fund Price Consumer address that is used by the pool
+    function rariFundPriceConsumer() external view returns (IRariFundPriceConsumer);
+
+    /// @return The pool's Yield Bearing Token (Fund Token)
+    function rariFundToken() external view returns (address);
+
+    /// @return An array of the symbols of the currencies supported by the pool
+    function getAcceptedCurrencies() external view returns (string[] memory);
+
+    /// @return Withdrawl Fee Rate (in 18 decimal precision)
```
axic, line 33:
```suggestion
    /// @return Withdrawal Fee Rate (in 18 decimal precision)
```

contracts/protocols/rari/IRariFundManager.sol
```diff
@@ -0,0 +1,35 @@
...
+
+import "./IRariFundPriceConsumer.sol";
+
+/***
+    @notice based on https://github.com/Rari-Capital/rari-stable-pool-contracts/blob/386aa8811e7f12c2908066ae17af923758503739/contracts/RariFundManager.sol
+ */
+interface IRariFundManager {
+    /// @dev Deposits an `amount` of Backing Tokens into pool
+    /// @param amount The amount of Backing Tokens to be deposited
+    /// @param currencyCode The symbol of the token to be deposited
```
axic, line 12: Swap the order of these two?

contracts/protocols/rari/IRariFundManager.sol
```diff
@@ -0,0 +1,35 @@
...
+    function deposit(string calldata currencyCode, uint256 amount) external;
+
+    /// @dev Withdraws an `amount` of Backing Tokens from the pool
+    /// @param amount The amount of Backing Tokens to withdraw
+    /// @param currencyCode The symbol of the token to withdraw
```
axic, line 17: Same here.

contracts/pools/RariTempusPool.sol
axic: Looking at this function again, and it is kind of worrying. Basically `rariFundManager.withdraw` will withdraw *some* amount of tokens and return a value of how supposedly how much. Then we transfer them and record the real transfer in case that transfer occurred a fee.

Yeah, I think the best would be recording how much Rari actually transferred to the pool on withdrawal, instead of relying on this number.

contracts/pools/RariTempusPool.sol
yuval-weiss: Yeah that's what I wanted to do initially but we have this - https://github.com/tempus-finance/tempus-protocol/blob/master/contracts/TempusPool.sol#L88

If we would add the `require` statement inside `calculateInterestRate` it would mean that:
1. When deploying the contract, the same require statement will be checked twice, in 2 different places (in `calculateInterestRate` and in `TempusPool::constructor`.
2. [This unit test](https://github.com/tempus-finance/tempus-protocol/blob/master/test/TempusPool.Deploy.test.ts#L85) fails

contracts/protocols/rari/IRariFundManager.sol
yuval-weiss: yup

contracts/protocols/rari/IRariFundPriceConsumer.sol
yuval-weiss: Sure, added :)

contracts/protocols/rari/IRariFundPriceConsumer.sol
yuval-weiss: Right. Done (all over the PR, not just this specific case)

contracts/pools/RariTempusPool.sol
yuval-weiss: I completely agree. Done.

contracts/mocks/rari/RariFundManagerMock.sol
```diff
@@ -0,0 +1,161 @@
...
+
+        _initialFundBalance = 1000 * (10**_asset.decimals()); /// 1000 BackingTokens
```
axic, line 47: Hmm, why is there a need for this? The other mocks don't start with an initial holding.

test/utils/PoolShare.ts
yuval-weiss: Done

contracts/mocks/rari/RariFundManagerMock.sol
yuval-weiss: Yeah I had some issues when I started working on this mock and for some reason, I thought it's a good idea to do this. I agree that it's redundant and so I deleted it and refactored some stuff so it'll still work

test/utils/PoolShare.ts
```diff
@@ -1,4 +1,4 @@
-import { Contract } from "ethers";
+import { Contract, Transaction } from "ethers";
```
axic, line 1: Not needed now.

test/pool-utils/RariTestPool.ts
```diff
@@ -0,0 +1,46 @@
...
+  constructor(ASSET_TOKEN:TokenInfo, YIELD_TOKEN:TokenInfo, integration:boolean) {
+    super(PoolType.Rari, /*acceptsEther*/false, /*yieldPeggedToAsset:*/false, integration);
+    this.ASSET_TOKEN = ASSET_TOKEN;
+    this.YIELD_TOKEN = YIELD_TOKEN;
+  }
+  async setInterestRate(rate:number): Promise<void> {
+    await this.rari.setInterestRate(rate);
+    /// TODO: temporary hack that updates the cached interest rate. Should be removed once
+    ///   Stats.sol exposes non-view methods that use the latest rate (then tests could be updated to use these).
+    await this.yields.contract.getPricePerFullShare();
```
axic, line 21: Hmm, actually would this "hack" be needed in other pool tests, or this specific to rari's mock?

test/pool-utils/RariTestPool.ts
```diff
@@ -0,0 +1,46 @@
...
+    /// TODO: temporary hack that updates the cached interest rate. Should be removed once
+    ///   Stats.sol exposes non-view methods that use the latest rate (then tests could be updated to use these).
+    await this.yields.contract.getPricePerFullShare(); 
+  }
+  async forceFailNextDepositOrRedeem(): Promise<void> {
+    await this.rari.contract.setFailNextDepositOrRedeem(true);
+  }
+  async getSigners(): Promise<[Signer,Signer,Signer]> {
+    if (this.integration) {
+      // TODO: implement for integration tests
```
axic, line 28: Hmm?

test/utils/RariFundManager.ts
```diff
@@ -0,0 +1,43 @@
...
+  static async create(ASSET:TokenInfo, YIELD:TokenInfo, initialRate:number = 1.0): Promise<RariFundManager> {
+    const asset = await ERC20.deploy("ERC20FixedSupply", ASSET.decimals, ASSET.decimals, ASSET.name, ASSET.symbol, parseDecimal(ASSET.totalSupply, ASSET.decimals));
+    const rariFundManager = await ContractBase.deployContract(
+      "RariFundManagerMock",
+      asset.address,
+      parseDecimal(initialRate, 18), // rate is always 18 decimals
+      YIELD.name,
+      YIELD.symbol
+    );
+    const yieldToken = await ERC20.attach("ERC20FixedSupply", (await rariFundManager.rariFundToken()), YIELD.decimals);
```
axic, line 31: I think could/should have been `ERC20` only here and in other mocks, but it doesn't matter for this PR, so lets keep as-is and refactor/change later.

contracts/protocols/rari/IRariFundPriceConsumer.sol
axic: Can you also mention this list of supported currencies come and how it is ordered?

Is this just a hardcoded list from Rari?

contracts/protocols/rari/IRariFundPriceConsumer.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity 0.8.10;
+
+/// @notice based on https://github.com/Rari-Capital/rari-stable-pool-contracts/blob/386aa8811e7f12c2908066ae17af923758503739/contracts/RariFundPriceConsumer.sol
+interface IRariFundPriceConsumer {
+    /// @return the price of each supported currency in USD (scaled by 1e18).
+    function getCurrencyPricesInUsd() external view returns (uint256[] memory);
```
axic, line 7:
```suggestion
    /// `IRariFundManager.getAcceptedCurrencies()` returns the supported currency symbols.
    /// Each `IRariFundManager` has an associated `IRariFundPriceConsumer`, and the prices
    /// returned here correspond to those currencies, in the same order.
    function getCurrencyPricesInUsd() external view returns (uint256[] memory);
```

axic: Not sure if this true, but it is my impression ^^

test/pool-utils/RariTestPool.ts
yuval-weiss: It's specific to Rari. Not because of the rari mock but because the `RariTempusPool` caches the storedInterestRate internally. This `getPricePerFullShare` call is made to trigger `updateInterestRate`, which will in turn update the cached `storedInterestRate`.

test/utils/RariFundManager.ts
yuval-weiss: Yeah I agree. Closing this comment then.

test/pool-utils/RariTestPool.ts
yuval-weiss: I guess it's a bit unclear but we have 2 different types of integration tests:
1. Integration tests under `/integration-tests`. (`yarn test:integration`)
2. Unit tests under `/test` can be run as "integration tests". (`yarn test:integration:lido`)

This if statement is related to the second type because we only have support for Lido atm.

contracts/mocks/rari/RariFundManagerMock.sol
```diff
@@ -0,0 +1,154 @@
...
+        uint256 unscaledDeposits = (asset.balanceOf(address(this)) - _backingTokensDeposited).mulfV(
+            backingTokenToUsdRate,
+            10**asset.decimals()
+        );
+
+        return scaledFundBalance + unscaledDeposits;
+    }
+
+    function getWithdrawalFeeRate() external pure override returns (uint256) {
+        return 0;
```
axic, line 79: Do they have this hardcoded as 0? If not, thsi should be modifyable (like `setInterestRate`) and tests should look at this case.

yuval-weiss: non-zero fee is tested in integration test... It's a protocol-specific thing so I don't think it should be included in the TempusPool unit tests

contracts/pools/RariTempusPool.sol
yuval-weiss: Closing this...

contracts/pools/RariTempusPool.sol
yuval-weiss: Closing this....

test/pool-utils/RariTestPool.ts
yuval-weiss: Closing this...

test/pool-utils/RariTestPool.ts
yuval-weiss: Closing this....

contracts/protocols/rari/IRariFundPriceConsumer.sol
yuval-weiss: Yup, it's hardcoded. Added a note + ref for that.

contracts/pools/RariTempusPool.sol
yuval-weiss: Closing dis.

contracts/pools/RariTempusPool.sol
yuval-weiss: Closing this, feel free to re-open if you have any more comments

contracts/protocols/rari/IRariFundPriceConsumer.sol
yuval-weiss: lgtm

contracts/mocks/rari/RariFundManagerMock.sol
axic: But the integration tests also only test it with 0, don't they? :)

axic: Basically the worry is that in case the integration is broken with non-zero fee, we have no idea. I know you've mentioend that upstream promised not to turn on fees again, but is only a soft commitment.

yuval-weiss: Each integration test case is tested twice - once with 0 fee and once with a non-zero fee...

test/pool-utils/RariTestPool.ts
axic: Can you document this there please?

yuval-weiss: Sure, done.

contracts/mocks/rari/RariFundManagerMock.sol
RedFox20: I think we should make this into a task and look back into it later. I can help with this after Rari is merged. Is that ok @axic ? I just want us to move forward a bit and split remaining work into subtasks between backend team 🙂

RedFox20: https://github.com/tempus-finance/tempus-protocol/issues/475

test/pool-utils/MultiPoolTestSuite.ts
```diff
@@ -56,6 +57,7 @@ function _describeForEachPoolType(title:string, poolTypes:PoolType[], only:boole
           case PoolType.Lido:     pool = new LidoTestPool(asset, yieldToken, integration); break;
           case PoolType.Compound: pool = new CompoundTestPool(asset, yieldToken, integration); break;
           case PoolType.Yearn:    pool = new YearnTestPool(asset, yieldToken, integration); break;
+          case PoolType.Rari:    pool = new RariTestPool(asset, yieldToken, integration); break;
```
RedFox20, line 60: I'll try to make this pool instancing automated. Kinda lame that we do this by hand, adds unnecessary maintenance.
https://github.com/tempus-finance/tempus-protocol/issues/479

test/utils/TempusPool.ts
```diff
@@ -292,6 +324,30 @@ export class TempusPool extends ContractBase {
           matureRedeemPercent: yieldToken.toBigNum(0.5)
         }
       );
+    } else if (type === PoolType.Rari) {
```
RedFox20, line 327: https://github.com/tempus-finance/tempus-protocol/issues/478

test/utils/RariFundManager.ts
```diff
@@ -0,0 +1,43 @@
+import { Contract } from "ethers";
+import { NumberOrString, parseDecimal } from "./Decimal";
+import { ContractBase, SignerOrAddress } from "./ContractBase";
+import { ERC20 } from "./ERC20";
+import { TokenInfo } from "test/pool-utils/TokenInfo";
+
+export class RariFundManager extends ContractBase {
```
RedFox20, line 7: Implement RariFundManagerFork
https://github.com/tempus-finance/tempus-protocol/issues/477

test/pool-utils/RariTestPool.ts
```diff
@@ -0,0 +1,47 @@
...
+    await this.rari.setInterestRate(rate);
+    /// TODO: temporary hack - this getPricePerFullShare call is made to trigger updateInterestRate,
+    ///     which will in turn update the cached storedInterestRate
+    /// Should be removed once Stats.sol exposes non-view methods that use the latest rate (then tests could be updated to use these).
+    await this.yields.contract.getPricePerFullShare(); 
+  }
+  async forceFailNextDepositOrRedeem(): Promise<void> {
+    await this.rari.contract.setFailNextDepositOrRedeem(true);
+  }
+  async getSigners(): Promise<[Signer,Signer,Signer]> {
```
RedFox20, line 27: Implement signers for integration tests
https://github.com/tempus-finance/tempus-protocol/issues/476

test/TempusPool.DepositBackingTokens.test.ts
```diff
@@ -80,7 +80,8 @@ describeForEachPool("TempusPool DepositBackingTokens", (pool:PoolTestFixture) =>
     (await pool.userState(user)).expect(depositAmount, depositAmount, /*yieldBearing:*/500);
   });
 
-  it.includeIntegration("Should issue appropriate shares after depositing Backing Tokens after changing rate to 2.0", async () =>
+  it("Should issue appropriate shares after depositing Backing Tokens after changing rate to 2.0", async () =>
+  // it.includeIntegration("Should issue appropriate shares after depositing Backing Tokens after changing rate to 2.0", async () =>
```
RedFox20, line 84: @yuval-weiss Why was this commented out again?

integration-tests/TempusPool.yearn.test.ts
RedFox20: https://github.com/tempus-finance/tempus-protocol/issues/480

integration-tests/TempusPool.rari.test.ts
```diff
@@ -0,0 +1,197 @@
...
+import { generateTempusSharesNames, TempusPool } from "../test/utils/TempusPool";
+import { ERC20 } from "../test/utils/ERC20";
+import { TempusController } from "../test/utils/TempusController";
+import { fromWei, parseDecimal, formatDecimal, toWei } from "../test/utils/Decimal";
+import Decimal from "decimal.js";
+import { BigNumberish } from "@ethersproject/bignumber";
+
+
+const fixtures = {} // maps RariWithdrawalFee --> fixture 
+async function setupWithRariWithdrawalFee(rariFee: BigNumberish) {
```
RedFox20, line 18: https://github.com/tempus-finance/tempus-protocol/issues/481

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,177 @@
...
+            fundManager.rariFundToken(),
+            backingToken,
+            controller,
+            maturity,
+            calculateInterestRate(
+                fundManager,
+                fundManager.rariFundToken(),
+                getTokenRariPoolIndex(fundManager, backingToken)
+            ),
+            /*exchangeRateOne:*/
```
axic, line 45: Ugh, linter made this real nice by breaking it up lol.

test/TempusPool.DepositBackingTokens.test.ts
yuval-weiss: Whoops, good catch

contracts/pools/RariTempusPool.sol
axic: Moved to #483.

contracts/pools/RariTempusPool.sol
```diff
@@ -0,0 +1,177 @@
...
+    function numYieldTokensPerAsset(uint backingTokens, uint rate) public view override returns (uint) {
+        return backingTokens.divfV(rate, exchangeRateONE) * exchangeRateToBackingPrecision;
+    }
+
+    /// @dev The rate precision is always 18
+    function interestRateToSharePrice(uint interestRate) internal view override returns (uint) {
+        return interestRate / exchangeRateToBackingPrecision;
+    }
+
+    /// @notice - based on https://github.com/Rari-Capital/rari-stable-pool-contracts/blob/386aa8811e7f12c2908066ae17af923758503739/contracts/RariFundManager.sol#L580
```
axic, line 142:
```suggestion
    /// We need to duplicate this, because the Rari protocol does not expose it.
    /// 
    /// Base on https://github.com/Rari-Capital/rari-stable-pool-contracts/blob/386aa8811e7f12c2908066ae17af923758503739/contracts/RariFundManager.sol#L580
```

RedFox20 (approved):
I think the initial version is OK. We should continue with finishing Rari sub tasks

PR #445, opened 2021-12-02
test: Add redeem test for fluctuating interest rate and rework redemption tests

axic:


test/TempusPool.Redeem.test.ts
```diff
@@ -179,17 +179,106 @@ describeForEachPool("TempusPool Redeem", (pool:PoolTestFixture) =>
...
+    }
+    else
+    {
+        (await pool.userState(user)).expect(90, 90, /*yieldBearing:*/110);
+    }
+
+    await pool.setTimeRelativeToPoolStart(0.3); // 30% of the time, i.e. ~5 days
+    await pool.setInterestRate(1);
+
+    await pool.redeemToYBT(user, 10, 10);
```
axic, line 231: Moving these into a new helper called `redeemAndCheckYBT`.

test/TempusPool.Redeem.test.ts
```diff
@@ -2,9 +2,32 @@ import { expect } from "chai";
 import { PoolTestFixture } from "./pool-utils/PoolTestFixture";
 import { describeForEachPool, integrationExclusiveIt as it } from "./pool-utils/MultiPoolTestSuite";
 import { expectRevert } from "./utils/Utils";
+import { Signer } from "./utils/ContractBase";
 
 describeForEachPool("TempusPool Redeem", (pool:PoolTestFixture) =>
 {
+  interface RedemptionExpectation {
+    redeemedTPS: number,
```
axic, line 10: And alternative naming scheme I considered was `tpsRedeemed` and `tpsBalance`, etc.

axic: These need to be low-level `number` and not `Number` because some functions expect that.

test/TempusPool.Redeem.test.ts
```diff
@@ -150,44 +161,51 @@ describeForEachPool("TempusPool Redeem", (pool:PoolTestFixture) =>
     await pool.depositYBT(user, 100, /*recipient:*/user);
     (await pool.userState(user)).expect(100, 100, /*yieldBearing:*/100);
 
-    await pool.setTimeRelativeToPoolStart(0.125); // 12.5% of the time, i.e. 1 day
+    await pool.setTimeDaysAfterPoolStart(1);
     await pool.setInterestRate(0.9);
+    await redeemAndCheckYBT(user, { redeem: { tps: 10, tys: 10 }, balanceAfter: { tps: 90, tys: 90 }, ybtAfter: { pegged: 99, unpegged: 110 } }, "10 shares at 0.9 yield should be 9");
```
axic, line 166: @RedFox20 so this one has the expectation message, not sure how much it really helps, but the lines are kind of getting too wide

axic:
@RedFox20 this has all the changes we discussed, expect didn't add yet all the expectation messages. I'm fine with either of the helpers. If you really want the expectation messages, some help with that would be appreciated. I also want to squash this before merge, but I think it is pretty close now.

test/TempusPool.Redeem.test.ts
RedFox20: Yeah, this is good for now, we can split this up over multiple lines later

RedFox20:
> @RedFox20 this has all the changes we discussed, expect didn't add yet all the expectation messages. I'm fine with either of the helpers. If you really want the expectation messages, some help with that would be appreciated. I also want to squash this before merge, but I think it is pretty close now.

Hmm, I think this would be like 15 minutes of work max, don't see what I could help with here, it should be very straighforward

RedFox20 (approved):

axic:
I can duplicate the above pattern for the negative yield cases, but I am not entirely sure what expectation messages you want to add, that's all :)

RedFox20:
Oh it's relatively simple, you just write what your intent is. For example "redeem 100 TPS+TYS expect all shares burned and receive YBT"

RedFox20 (approved):

PR #446, opened 2021-12-02
Make TempusPool.updateInterestRate public

yuval-weiss:


contracts/ITempusPool.sol
```diff
@@ -196,6 +196,11 @@ interface ITempusPool is ITempusFees {
         bool toBackingToken
     ) external view returns (uint256);
 
+    /// @dev This updates the underlying pool's interest rate
+    ///      It should be done first thing before deposit/redeem to avoid arbitrage
```
axic, line 200: Done by whom? Deposit/redeem should already call this.

RedFox20: Yeah this doc is outdated. We should just say this is always done first thing before deposit/redeem

RedFox20 (changes requested):
Only small fix in docs needed

contracts/ITempusPool.sol
mijovic, line 200: I would suggest something like this
```suggestion
    ///      It is done first thing before deposit/redeem to avoid arbitrage
    ///      It is available to call publically to periodically update interest rates in cases of low volume
```

mijovic (approved):

RedFox20 (approved):

PR #447, opened 2021-12-03
Fix backing token for local deploy

lazarmitic:
I'm not sure how anything worked with backing token address set to ETH, but the actual backing token of the pool was WETH. I was able to mint using ETH inside WETH pool 😕

### DIFF
--- a/test/utils/ERC20Ether.ts
+++ b/test/utils/ERC20Ether.ts
@@ -8,6 +8,9 @@ import { IERC20 } from "./IERC20";
  * so that we can simplify tests on native ETH based pools
  */
 export class ERC20Ether extends DecimalConvertible implements IERC20 {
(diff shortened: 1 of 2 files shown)


test/utils/ERC20Ether.ts
```diff
@@ -8,6 +8,9 @@ import { IERC20 } from "./IERC20";
  * so that we can simplify tests on native ETH based pools
  */
 export class ERC20Ether extends DecimalConvertible implements IERC20 {
+  // Address of this contract - ETH is Zero address
+  address = '0x0000000000000000000000000000000000000000';
```
RedFox20, line 12: Nice, thanks for adding this 😃

RedFox20 (approved):

PR #449, opened 2021-12-03
pool: Make maximumNegativeYieldDuration configurable

axic:


axic:
Overriding this setting in each pool implementation is also a good option. The only question we need to be confident on: does the likelyhood/acceptability of negative yield depend on 1) protocol only or 2) the asset can be a differentiator too.

If the answer is 1) then overriding this setting in the pool implementation is the better choice. If the answer is 2) then this PR is the better.

PR #450, opened 2021-12-03
controller: Add version constant

axic:


RedFox20:
Lets use a version string instead like Yuval's new PR with string constant "1.0.0"

axic:
Can you explain the reasoning behind a version string?

yuval-weiss:
This can be closed now I guess

PR #451, opened 2021-12-03
added versioning to TempusController & Stats; changed version type to string

yuval-weiss:


RedFox20:
Looks good, I think you need small update to TempusPool.ts as well for completeness sake

contracts/TempusController.sol
```diff
@@ -21,6 +21,8 @@ contract TempusController is ReentrancyGuard, Ownable {
     using UntrustedERC20 for IERC20;
     using AMMBalancesHelper for uint256[];
 
+    string public constant version = "1.0.0";
```
axic, line 24: String is not a very good choice for version numbers, should use semver. Semver defines what exact number means, ie. on major=1 any change to minor is fine and non breaking, however programatically finding this out means a lot of text parsing, which is not ideal off-chain (but doable) and highly inefficient on-chain.

axic: A better option is the usual way to combine these into a single number: `(major << 16) | (minor << 8) | patch`.

contracts/stats/Stats.sol
```diff
@@ -15,6 +15,8 @@ contract Stats is ITokenPairPriceFeed, ChainlinkTokenPairPriceFeed {
     using Fixed256xVar for uint256;
     using AMMBalancesHelper for uint256[];
 
+    string public constant version = "1.0.0";
```
RedFox20, line 18: What about Stats contract?

test/TempusPool.Deploy.test.ts
```diff
@@ -16,7 +16,7 @@ describeForEachPool("TempusPool Deploy", (testPool:PoolTestFixture) =>
 
   it("Version is correct", async () =>
   {
-    expect(await pool.version()).to.equal(1);
+    expect(await pool.version()).to.equal("1.0.0");
```
RedFox20, line 19: I guess this needs an update now?

yuval-weiss: yup, done.

test/TempusController.test.ts
```diff
@@ -50,6 +50,21 @@ describeForEachPool("TempusController", (testPool:PoolTestFixture) =>
     expect(+await pool.yieldShare.balanceOf(controller.address)).to.be.lessThan(2e-18, "No funds should remain in controller");
   }
 
+  describe("deploy", async () => {
+    it("Version is correct", async () =>
+  {
```
RedFox20, line 55: Some whitespace issues here, but otherwise looks great

contracts/stats/Stats.sol
yuval-weiss: What about it? I added versioning to Stats as well..

test/stats/Stats.test.ts
```diff
@@ -32,6 +32,14 @@ describeForEachPool("Stats", (testPool:PoolTestFixture) =>
     stats = await Stats.create();
   });
 
+  it("Version is correct", async () =>
+  {
+    const { major, minor, patch } = await stats.version();
```
RedFox20, line 37: That's what's kind of annoying here, would be easy to just check for "1.0.0" instead

RedFox20 (approved):

PR #452, opened 2021-12-06
Fix stETH address casing

lazarmitic:


RedFox20:
@lazarmitic Did this cause some issues on frontend? It is quite common to have mixed casing for readability, so I suggest preprocessing all addresses with .toLower()

lazarmitic:
Casing of this address and address contained in transaction receipt for stETH did not match. This change updates address to have same casing as when you call `. address()` for stETH.

Yeah, I can call to lower case, but I tried to avoid doing that

lazarmitic:
Original address was lower case, one found on Etherscan is mixed casing and I want to set that, just to be consistent with address casing you get when fetching stETH address from ERC20 contract.

axic:
> It is quite common to have mixed casing for readability

The mixed-case is not for readability in the context of Ethereum, but it is the [ERC-55 style checksum](https://eips.ethereum.org/EIPS/eip-55). Some tools heavily rely on it / enforce it (even Solidity source code!).  In some cases even ethers.js requires it.

RedFox20:
> > It is quite common to have mixed casing for readability
> 
> The mixed-case is not for readability in the context of Ethereum, but it is the [ERC-55 style checksum](https://eips.ethereum.org/EIPS/eip-55). Some tools heavily rely on it / enforce it (even Solidity source code!). In some cases even ethers.js requires it.

Oh damn, why would they do that? Then how should we handle input addresses?

Although Lazar's comment about this being an internal issue makes sense, so I'm inclined to just merge this

axic:
> Then how should we handle input addresses?

Input addresses where exactly?

RedFox20:
> Input addresses where exactly?

I mean where-ever we have frontend deal with these addresses, maybe @lazarmitic can check if there are any potential issues. It concerns me that he had to make this change here in order to fix something?

axic:
> > Input addresses where exactly?
> 
> I mean where-ever we have frontend deal with these addresses, maybe @lazarmitic can check if there are any potential issues. It concerns me that he had to make this change here in order to fix something?

Yes, if there is a user entered address in a text box, this can become a question -- we should have some validation of that to avoid mistakes.

lazarmitic:
I will provider bigger context here:

1. When user executes a transaction (let's say redeem) we want to show notification that contains how much TPS and TYS were swapper for stETH.
2. In order to do above, I get transaction receipt.
3. Receipt contains a bunch of token transfers and transfer amounts.
4. I need to find stETH transfer in this array of transactions inside receipt by iterating over them and looking for stETH transaction based on stETH address.
5. This part here fails because stETHConfigAddress !== stETHContract address (they are same but different casing).

Change in this PR fixes above problem.

Another solution is to call .toLowerCase every time we compare addresses, but I think it's much easier to just use same casing for address that we get from Etherscan or token ERC20 contract.

lazarmitic:
Currently we do not have any text boxes in which users enter token addresses.

axic:
> I need to find stETH transfer in this array of transactions inside receipt by iterating over them and looking for stETH transaction based on stETH address.

There you could definitely work with `.toLower()`, but any user input should be validated according to the checksum.

RedFox20:
@lazarmitic In this case, we can merge this PR, but the frontend should definitely normalize hashes before comparison to future-proof it.

lazarmitic:
Okey, I agree, I will update frontend as well to make it more robust.

RedFox20 (approved):

Issue #173, opened 2021-08-19
Implement more tests to reach ~100% coverage


mijovic:
- [ ] Fees in TempusAMM
- [ ] `AavePool.depositToUnderlying` with value
- [ ] `CompoundPool.depositToUnderlying` with value
- [ ] Sanity checks in `CompoundPool`
- [ ] `LidoPool.depositToUnderlying` without value
- [ ] `LidoPool.withdrawFromUnderlying`

PR #453, opened 2021-12-06
test: Add missing test cases for deposit/redeem to 0 address

axic:
Part of #173.

### DIFF
--- a/integration-tests/TempusPool.compound.test.ts
+++ b/integration-tests/TempusPool.compound.test.ts
@@ -124,7 +124,7 @@ describe('TempusPool <> Compound <> DAI', function () {
 
     await evmSetAutomine(false);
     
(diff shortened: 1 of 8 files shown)


integration-tests/TempusPool.compound.test.ts
```diff
@@ -124,7 +124,7 @@ describe('TempusPool <> Compound <> DAI', function () {
 
     await evmSetAutomine(false);
     
-    await tempusPool.controller.redeemToBacking(signer1, tempusPool, yieldShareBalanceSigner1, yieldShareBalanceSigner1);
+    await tempusPool.controller.redeemToBacking(signer1, tempusPool, yieldShareBalanceSigner1, yieldShareBalanceSigner1, signer1.address);
```
axic, line 127: Interestingly this file had another case of `redeemToBacking` which had the recipient, but because of TS that was ignored.

RedFox20 (approved):

PR #454, opened 2021-12-07
Refactor test utils for deposit and redeem

RedFox20:


test/pool-utils/PoolTestFixture.ts
```diff
@@ -13,6 +13,41 @@ import { TempusAMM } from "../utils/TempusAMM";
...
+}
+
+export interface DepositExpectation {
+  amount:DepositAmounts,
+  balanceAfter:ShareExpectation,
+  ybtAfter:YBTExpectation
+}
+
+export interface RedemptionExpectation {
+  redeem:ShareExpectation,
```
axic: Nitpick: should rename this to `amount`? Or `amount` to `deposit` in `DepositExpectation`.

test/pool-utils/PoolTestFixture.ts
```diff
@@ -333,6 +368,48 @@ export abstract class PoolTestFixture {
...
+  /**
+   * TESTING UTILITY checks user state for TPS+TYS balance and YBT balance
+   * @param user User whose wallet to check
+   * @param shares Amount of TPS and TYS we expect
+   * @param ybt Amount of YBT we expect
+   * @param message Description of what we expected to happen
+   */
+  async checkWallet(user:Signer, shares:ShareExpectation, ybt:YBTExpectation, message?:string): Promise<void> {
+    const us = await this.userState(user);
+    us.expectMulti(shares.tps, shares.tys, ybt.pegged, ybt.pegged, message);
```
axic: Hmm, is this correct or wrong?

```suggestion
    us.expectMulti(shares.tps, shares.tys, ybt.pegged, ybt.unpegged, message);
```

test/pool-utils/PoolTestFixture.ts
```diff
@@ -333,6 +368,48 @@ export abstract class PoolTestFixture {
...
+  /**
+   * TESTING UTILITY: does a depositBT or depositYBT and then validates
+   *                  user wallet balances
+   * @param user User who is depositing and receiving shares
+   * @param deposit All the deposit and checks parameters
+   * @param message Description of what we expected to happen
+   */
+  async depositAndCheck(user:Signer, deposit:DepositExpectation, message?:string): Promise<void>
+  {
+    if (deposit.amount.bt > 0) {
```
axic: Hm for the sake of wrong tests, I think this should have an assert for `ybt == 0`, and the other case for `bt == 0`.

test/TempusPool.Redeem.test.ts
```diff
@@ -83,9 +83,10 @@ describeForEachPool("TempusPool Redeem", (pool:PoolTestFixture) =>
     await pool.createDefault();
     let [owner, user] = pool.signers;
     await pool.setupAccounts(owner, [[user, 100]]);
-
-    await pool.depositYBT(user, 100, /*recipient:*/user);
-    (await pool.userState(user)).expect(100, 100, /*yieldBearing:*/0);
+    await pool.depositAndCheck(user,
+      { deposit:YBT(100), balanceAfter:{tps:100, tys:100}, ybtAfter:{pegged:0, unpegged:0} },
```
axic, line 86: Nit: I'm not a big fan of leaving no space after semicolons, but probably the tests use a wide mixture today.

test/pool-utils/PoolTestFixture.ts
```diff
@@ -14,30 +14,35 @@ import { PoolShare } from "../utils/PoolShare";
 import { strict as assert } from 'assert';
 
 export interface DepositAmounts {
-  bt:number,
-  ybt:number
+  bt:number;
+  ybt:number;
```
axic, line 18: It seems semicolons is correct, but why did this accept colons previously? 😕

test/pool-utils/PoolTestFixture.ts
axic: This is what I get when I review commit-by-commit.

test/pool-utils/PoolTestFixture.ts
axic: This still stands after all the commits.

test/pool-utils/PoolTestFixture.ts
RedFox20: Yeah, I made a mistake in the original commit, was fixed in the next commit

test/TempusPool.Redeem.test.ts
RedFox20: It was getting really hard to fit these lines on my laptop, so I tried to compact it a little bit, hopefully it won't impact readability

test/pool-utils/PoolTestFixture.ts
RedFox20: Maybe TS supports commas in statements like this: `bt:number, ybt:number`
I would just use semicolons though

test/pool-utils/PoolTestFixture.ts
RedFox20: I will add an assert for this

test/pool-utils/PoolTestFixture.ts
```diff
@@ -333,6 +373,48 @@ export abstract class PoolTestFixture {
...
+   * @param deposit All the deposit and checks parameters
+   * @param message Description of what we expected to happen
+   */
+  async depositAndCheck(user:Signer, deposit:DepositExpectation, message?:string): Promise<void>
+  {
+    if (deposit.deposit.bt > 0) {
+      assert.ok(deposit.deposit.ybt === 0, "if BT > 0, then YBT must be 0");
+      await this.depositBT(user, deposit.deposit.bt);
+    } else {
+      await this.depositYBT(user, deposit.deposit.ybt);
```
axic, line 400: This still misses the `assert(deposit.deposit.bt == 0)`.  It could be negative, but more for symmetric correctness.

RedFox20: I thought about this and it happens that the first branch already handles the case where both BT and YBT are set, so it can never be true in the second branch.

axic:
Since a lot of these commits were not self contained, but subsequent commits fixed bugs in previously introduced ones, I would suggest to squash the entire thing into a single one as they can't be independently reviewed anyway.

test/pool-utils/PoolTestFixture.ts
RedFox20: Hmm, yeah, those would cause everything to explode though. So should we add complete number validation here somewhere do you think? It would be cheap on the TypeScript side

axic: The ideal would be a "number range" type. That is what I'd had needed in one test case where it has to be `>0`. For the matter of this PR I would just add the assertion and perhaps consider working on a generic number range type separately.

axic: For the test case I mentioned just grep for `assert.ok` and you should find it.

RedFox20: Refactored, check the new solution

test/pool-utils/PoolTestFixture.ts
```diff
@@ -333,6 +368,48 @@ export abstract class PoolTestFixture {
...
+   * @param message Description of what we expected to happen
+   */
+  async depositAndCheck(user:Signer, deposit:DepositExpectation, message?:string): Promise<void>
+  {
+    if (deposit.amount.bt > 0) {
+      await this.depositBT(user, deposit.amount.bt);
+    } else {
+      await this.depositYBT(user, deposit.amount.ybt);
+    }
+    await this.checkWallet(user, deposit.balanceAfter, deposit.ybtAfter, message);
```
axic: Hmm, actually this function is a bit misleading as it only compares `ybtAfter` but either deposits BT or YBT.

RedFox20: Yeah, that's true, it's a bit confusing that way, someone deposits DAI and their cDAI balance shouldn't really change.

So maybe I should just replace `amount:YBT(100)` pattern to `ybtAmount:100` ? What do you think?

axic (approved):
Looks ok, but haven't checked every single test case replacement. The ones I checked were good.

PR #455, opened 2021-12-07
test: Add test case for depositing Ether when it is disallowed

axic:


test/TempusPool.Deposit.test.ts
```diff
@@ -65,6 +65,17 @@ describeForEachPool("TempusPool Deposit", (pool:PoolTestFixture) =>
...
   });
 
+  it("Should revert on Eth transfer as BT when not accepted", async () =>
+  {
+    await pool.createDefault();
+    const [owner] = pool.signers;
+
+    // If the backing token is not the zero address, then Ether transfers are not allowed
+    if ((await pool.tempus.backingToken()) !== '0x0000000000000000000000000000000000000000') {
+      (await expectRevert(pool.tempus.controller.depositBacking(owner, pool.tempus, 100, owner, /* ethValue */ 1))).to.equal('given TempusPool\'s Backing Token is not ETH');
```
RedFox20, line 75: I think PoolTestFixture.ts `depositBT()` and `expectDepositBT()` functions just needs an optional `ethValue:NumberOrString = 0` argument. And then this expression should be 
```ts
(await pool.expectDepositBT(owner, 100, /*ethValue*/ 1)).to.equal(''given TempusPool\'s Backing Token is not ETH");
```

test/pool-utils/PoolTestFixture.ts
```diff
@@ -176,8 +176,8 @@ export abstract class PoolTestFixture {
   /**
    * Deposit BackingTokens into TempusPool
    */
-  async depositBT(user:Signer, backingTokenAmount:NumberOrString, recipient:SignerOrAddress = user): Promise<Transaction> {
-    return this.tempus.controller.depositBacking(user, this.tempus, backingTokenAmount, recipient);
+  async depositBT(user:Signer, backingTokenAmount:NumberOrString, recipient:SignerOrAddress = user, ethValue: NumberOrString = 0): Promise<Transaction> {
```
RedFox20, line 179: Since we do an override in `LidoTestPool.ts`, what will happen to that? The last argument is silently ignored?

axic: Yes. Though this test is not affected because of the `backingToken != 0` condition.

axic: Do you have any strong opinion for not just going with the first commit and then arguing about the second later?

RedFox20 (approved):

RedFox20 (dismissed):

RedFox20 (approved):

PR #457, opened 2021-12-07
pool: Turn require statement for Eth deposits into assertion as it can't be triggered

axic:


contracts/pools/AaveTempusPool.sol
```diff
@@ -50,7 +50,8 @@ contract AaveTempusPool is TempusPool {
     }
 
     function depositToUnderlying(uint256 amount) internal override returns (uint256) {
-        require(msg.value == 0, "ETH deposits not supported");
+        // ETH deposits are not accepted, because it is rejected in the controller
+        assert(msg.value == 0);
```
RedFox20, line 54: I think then we don't need this check at all

RedFox20: Ok sure, if this overhead is minimal we can keep it. However we already validate it in the controller, so this is kind of unnecessary...

RedFox20 (approved):

PR #458, opened 2021-12-08
test: Add ethValue to depositBT wrappers

axic:


test/pool-utils/PoolTestFixture.ts
```diff
@@ -216,8 +216,8 @@ export abstract class PoolTestFixture {
   /**
    * Deposit BackingTokens into TempusPool
    */
-  async depositBT(user:Signer, backingTokenAmount:NumberOrString, recipient:SignerOrAddress = user): Promise<Transaction> {
-    return this.tempus.controller.depositBacking(user, this.tempus, backingTokenAmount, recipient);
+  async depositBT(user:Signer, backingTokenAmount:NumberOrString, recipient:SignerOrAddress = user, ethValue: NumberOrString = 0): Promise<Transaction> {
```
RedFox20, line 219: Repeating this comment from the previous PR. Maybe we should not override `depositBT` in LidoTestPool but instead do it explicitly here

axic: It would be nice doing the above, and it solves it for the general tests. However I think it will not make it possible to make those two special test cases to use it -- the once you are annoyed by directly using the underlying functions.

axic: Probably `(ethValue == undefined) ? backingTokenAmount : ethValue` is some kind of workaround.

RedFox20: No, this will not work. For most regular calls, attached ethereum must be 0.
The code snippet I provided, while inelegant, will do the trick.

We can replace Lido check with some sort of boolean like we do for pegged ybt-s yo make it nicer.

Lido test pool's `bt` field is an instance of ERC20Ether, which creates an ERC20-like wrapper around ETH transactions. We could check for that instance as well.

axic: > Probably (ethValue == undefined) ? backingTokenAmount : ethValue is some kind of workaround.

This is in addition to your snippet.

> The code snippet I provided, while inelegant, will do the trick.

With your snippet as-is, how can one forcefully send differing `backingTokenAmount` and `ethValue` to Lido?

RedFox20: Hmm, right, that is something I didn't think to cover yet. That would increase branch coverage since we didn't have such a test before. In this case for Lido we should have similar to what you proposed, but using JS-s `falsy` style instead with ||
```ts
  const eth = (this.type === PoolType.Lido) ? (ethValue || backingTokenAmount) : ethValue;
  return this.tempus.controller.depositBacking(user, this.tempus, backingTokenAmount, recipient, eth);
```

axic:
Need to clean up the commits, but I think this is getting there.

test/pool-utils/PoolTestFixture.ts
```diff
@@ -124,6 +124,9 @@ const POOL_FIXTURES: { [signature: string]: FixtureState } = {};
 export abstract class PoolTestFixture {
   type:PoolType;
 
+  // True if the pool accepts/requires Ether on deposits
+  acceptsEther:boolean;
```
RedFox20, line 128: Yeah, I think this is the only sane way to really do it without weird magic

test/pool-utils/PoolTestFixture.ts
```diff
@@ -216,8 +220,9 @@ export abstract class PoolTestFixture {
   /**
    * Deposit BackingTokens into TempusPool
    */
-  async depositBT(user:Signer, backingTokenAmount:NumberOrString, recipient:SignerOrAddress = user): Promise<Transaction> {
-    return this.tempus.controller.depositBacking(user, this.tempus, backingTokenAmount, recipient);
+  async depositBT(user:Signer, backingTokenAmount:NumberOrString, recipient:SignerOrAddress = user, ethValue: NumberOrString = 0): Promise<Transaction> {
+    const ethToTransfer = (this.type === PoolType.Lido) ? (ethValue || backingTokenAmount) : ethValue;
```
RedFox20, line 224: I think you forgot to upgrade this to `this.acceptsEther`

RedFox20 (approved):

PR #459, opened 2021-12-08
test: Add test for deposit mismatch for Lido

axic:


test/TempusPool.Deposit.test.ts
```diff
@@ -76,6 +76,20 @@ describeForEachPool("TempusPool Deposit", (pool:PoolTestFixture) =>
...
+  {
+    await pool.createDefault();
+    const [owner] = pool.signers;
+
+    // If the backing token is the zero address, then Ether transfers are allowed
+    if ((await pool.tempus.backingToken()) === '0x0000000000000000000000000000000000000000') {
+      // These two amounts are expected to be equal, but in this case they are deliberately different.
+      const depositAmount = 100;
+      const ethValue = 1;
+      (await expectRevert(pool.tempus.controller.depositBacking(owner, pool.tempus, depositAmount, owner, ethValue))).to.equal('ETH value does not match provided amount');
```
RedFox20, line 89: We shouldn't keep introducing technical debt with old-style deposits

axic: Yeah I agree, we should do this entire refactoring in #458.

RedFox20 (approved):

PR #460, opened 2021-12-08
test: Add proper tests for empty deposits

axic:


test/TempusPool.Deposit.test.ts
```diff
@@ -43,11 +43,18 @@ describeForEachPool("TempusPool Deposit", (pool:PoolTestFixture) =>
     );
   });
 
-  it.includeIntegration("Should revert on deposit 0", async () =>
+  it.includeIntegration("Should revert on depositing 0 BT", async () =>
```
RedFox20, line 46: BackingToken tests belong to TempusPool.DepositBackingToken.test.ts

axic: Hmm, actually there are many other BT-related cases, all of them should be moved. I would prefer to do that in a separate PR, maybe perhaps in #458.

RedFox20 (approved):

PR #462, opened 2021-12-09
test: refactor utilities for TempusPool.Redeem tests

RedFox20:


test/TempusPool.Redeem.test.ts
```diff
@@ -1,10 +1,32 @@
...
+    return pool.depositAndCheck(/*user*/pool.signers[1], expects, message);
+  }
+  
+  async function deposit2(user:Signer, expects:YBTDepositExpectation, message?:string): Promise<void> {
+    return pool.depositAndCheck(user, expects, message);
+  }
+
+  async function redeem(expects:YBTRedeemExpectation, message?:string): Promise<void> {
+    return pool.redeemAndCheck(/*user*/pool.signers[1], expects, message);
+  }
```
RedFox20, line 19: These are a little bit useless now, so I will delete them. I was working on multiple solutions to simplify these tests

test/TempusPool.Redeem.test.ts
```diff
@@ -143,288 +147,192 @@ describeForEachPool("TempusPool Redeem", (pool:PoolTestFixture) =>
...
+    await deposit2(user2, { ybtAmount:200, pegged:{tps:100, tys:100, ybt:800}, unpegged:{tps:200, tys:200, ybt:300} }, "deposit user2 200 with rate 2");
+
+    await pool.setInterestRate(2.5);
+    await pool.fastForwardToMaturity();
+    expect(await pool.tempus.initialInterestRate()).to.equal(1.0);
+    expect(await pool.tempus.currentInterestRate()).to.equal(2.5);
+    expect(await pool.tempus.maturityInterestRate()).to.equal(2.5);
+
+    // First user redeems
+    await pool.checkWallet(user, { pegged:{tps:150, tys:150, ybt:875}, unpegged:{tps:200, tys:200, ybt:300}}, "user1 pre-redeem");
```
RedFox20, line 320: So what happened here is that we have some tests where pegged/unpegged have different amounts of TPS/TYS, so this made all the utilities more complicated.

I decided to simply go with `pegged` and `unpegged` sets even if they have the same TPS/TYS. It might be possible to  do some magic with interfaces and support both options

test/pool-utils/PoolTestFixture.ts
```diff
@@ -60,32 +59,24 @@ export class UserState {
...
-    );
+    this.expect(principalShares, yieldShares, yieldBearing, message);
   }
 
-  // non-async to give us actual test failure line #
   public expect(principalShares:number, yieldShares:number, yieldBearing:number, message:string = null) {
-    const msg = message ? (": expected " + message) : "";
-    expect(this.principalShares).to.equal(principalShares, ("principalShares did not match expected value"+msg));
-    expect(this.yieldShares).to.equal(yieldShares, ("yieldShares did not match expected value"+msg));
+    const msg = (message||"") + " expected " + (this.yieldPeggedToAsset ? "pegged" : "unpegged")
```
RedFox20, line 68: I also improved error messages for expect(), this greatly helps with figuring out what went wrong.

test/pool-utils/PoolTestFixture.ts
```diff
@@ -374,38 +365,37 @@ export abstract class PoolTestFixture {
...
+   * TESTING UTILITY: checks user state for TPS+TYS balance and YBT balance
    * @param user User whose wallet to check
    * @param wallet All wallet check parameters
    * @param message Description of what we expected to happen
    */
   async checkWallet(user:Signer, wallet:WalletExpectation, message?:string): Promise<void> {
-    const us = await this.userState(user);
-    us.expectMulti(wallet.shares.tps, wallet.shares.tys, wallet.ybt.pegged, wallet.ybt.unpegged, message);
+    const expects:BalancesExpectation = this.yieldPeggedToAsset ? wallet.pegged : wallet.unpegged;
+    (await this.userState(user)).expect(expects.tps, expects.tys, expects.ybt, message);
```
RedFox20, line 375: One sad thing about calling `await this.userState(user)` here is that we lose stack traces. However, our traces are broken anyways right now 😞

test/TempusPool.Redeem.test.ts
mijovic: Hm, I am not sure if there is any magic that can help reducing complexity here, so it should be fine

test/pool-utils/PoolTestFixture.ts
mijovic: Nice

test/pool-utils/PoolTestFixture.ts
mijovic: Well, we can't do too much here. I don't mind stack traces being broken as well.

mijovic (approved):

PR #463, opened 2021-12-10
[exitTempusAmmAndRedeem] standardize minRate usage (always denominate rate as Yields rate)

yuval-weiss:


contracts/TempusController.sol
```diff
@@ -283,19 +279,32 @@ contract TempusController is ReentrancyGuard, Ownable, Versioned {
...
         uint256 maxLeftoverShares,
-        uint256 minRate,
+        uint256 yieldsRate,
+        uint256 maxSlippage,
         bool toBackingToken,
         uint256 deadline
     ) external nonReentrant {
         requireRegistered(address(tempusAMM));
-        _exitTempusAmmAndRedeem(
+        if (lpTokens > 0) {
```
axic, line 288: This purely seems to be a refactoring moving this from `_exitTempusAmmAndRedeem`. Is that the case or something else changed?

contracts/amm/interfaces/ITempusAMM.sol
```diff
@@ -39,11 +39,13 @@ interface ITempusAMM {
...
     /// @param threshold Maximum difference between final balances of Principals and Yields
     /// @return amountIn Amount of Principals or Yields that user needs to swap to end with almost equal amounts
+    /// @return yieldsIn Specifies the "direction" of the swap -
+    ///         whether Yields should be swapped for Principals (yieldsIn=true) or vice versa
     function getSwapAmountToEndWithEqualShares(
         uint256 principals,
         uint256 yields,
         uint256 threshold
-    ) external view returns (uint256 amountIn);
+    ) external view returns (uint256 amountIn, bool yieldsIn);
```
mijovic, line 48: Isn't this new return param redundant? It should be simple check `principals < yields`

mijovic: We can leave, just asking if it is really needed :)

contracts/TempusController.sol
```diff
@@ -225,8 +223,6 @@ contract TempusController is ReentrancyGuard, Ownable, Versioned {
     /// @dev Withdraws liquidity from TempusAMM and redeems Shares to Yield Bearing or Backing Tokens
     ///      Checks user's balance of principal shares and yield shares
     ///      and exits AMM with exact amounts needed for redemption.
-    /// @notice `msg.sender` needs to approve `tempusAMM.tempusPool` for both Yields and Principals
```
mijovic, line 228: we are burning shares so no need for approval here?

contracts/TempusController.sol
```diff
@@ -667,31 +675,34 @@ contract TempusController is ReentrancyGuard, Ownable, Versioned {
...
+                (uint256 swapAmount, bool yieldsIn) = tempusAMM.getSwapAmountToEndWithEqualShares(
+                    principals,
+                    yields,
+                    maxLeftoverShares
+                );
+                if (yieldsIn) {
+                    minReturn = swapAmount.mulfV(yieldsRate, tempusPool.backingTokenONE());
+                } else {
+                    minReturn = swapAmount.divfV(yieldsRate, tempusPool.backingTokenONE());
+                }
```
mijovic, line 697:
```suggestion
                uint256 minReturn = yieldsIn 
                    ? swapAmount.mulfV(yieldsRate, tempusPool.backingTokenONE())
                    : swapAmount.divfV(yieldsRate, tempusPool.backingTokenONE());
```

contracts/TempusController.sol
yuval-weiss: It was moved as a workaround to a "Stack too deep, try removing local variables" error that I was getting

contracts/amm/interfaces/ITempusAMM.sol
yuval-weiss: I think it makes sense to also return the direction of the swap from `getSwapAmountToEndWithEqualShares`... Otherwise all consumers of this function need to check `principals < yields` which is redundant as it's already checked internally in `getSwapAmountToEndWithEqualShares`

contracts/TempusController.sol
yuval-weiss: Yup

yuval-weiss: Users only need to approve the LP tokens

contracts/TempusController.sol
```diff
@@ -166,7 +166,6 @@ contract TempusController is ReentrancyGuard, Ownable, Versioned {
     }
 
     /// @dev Redeem TPS+TYS held by msg.sender into Yield Bearing Tokens
-    /// @notice `msg.sender` must approve Principals and Yields amounts to `targetPool`
```
RedFox20, line 169: No need to approve anything anymore?

mijovic: Yup, shares are burnt so no need to approve

contracts/TempusController.sol
```diff
@@ -667,31 +675,31 @@ contract TempusController is ReentrancyGuard, Ownable, Versioned {
...
-
-        if (lpTokens > 0) {
-            // if there is LP balance, transfer to controller
-            require(tempusAMM.transferFrom(msg.sender, address(this), lpTokens), "LP token transfer failed");
-
-            // exit amm and sent shares to controller
-            _exitTempusAMMGivenLP(tempusAMM, address(this), address(this), lpTokens, minLpAmountsOut, false);
-        }
+        require(yieldsRate > 0, "yieldsRate must be greater than 0");
+        require(maxSlippage <= 1e18, "maxSlippage must be less than 1e18");
```
mijovic, line 679:
```suggestion
        require(maxSlippage <= 1e18, "maxSlippage must be less than or equal 1e18");
```

mijovic: Or just change to `maxSlippage can not be greater than 1e18`

yuval-weiss: done

contracts/amm/interfaces/ITempusAMM.sol
RedFox20: Agree it is done in several places, so it makes sense to have it

test/TempusController.test.ts
```diff
@@ -314,6 +315,93 @@ describeForEachPool("TempusController", (testPool:PoolTestFixture) =>
...
+      await expect(controller.exitTempusAmmAndRedeem(
+        testPool, 
+        owner, 
+        "0", 
+        "0",
+        await testPool.yields.balanceOf(owner),
+        false,
+        currentYieldsRate.toString(),
+        "0.001"
+      )).to.be.revertedWith("BAL#507");
```
RedFox20, line 402:
```suggestion
      await expect(
        controller.exitTempusAmmAndRedeem(testPool, owner, 0, 0, await testPool.yields.balanceOf(owner), false, currentYieldsRate, 0.001)
      ).to.be.revertedWith("BAL#507");
```
If you reformat the code a bit, it could look much more readable in this way, at least in my opinion

test/TempusController.test.ts
```diff
@@ -314,6 +315,93 @@ describeForEachPool("TempusController", (testPool:PoolTestFixture) =>
...
+
+    it("Should fail swap due to minimum return Principals --> Yields w/ 0.1% Maximum Slippage", async () => 
+    {
+      await initAMM(user1, /*ybtDeposit*/1000000, /*principals*/100000, /*yields*/1000000);
+      await testPool.controller.depositYieldBearing(owner, testPool.tempus, 100);
+
+      const pricePerYield = await testPool.yields.getPricePerFullShareStored();
+      const pricePerPrincipal = await testPool.principals.getPricePerFullShareStored();
+      
+      const currentYieldsRate = new Decimal(pricePerYield.toString()).div(pricePerPrincipal.toString()).toFixed(testPool.principals.decimals);
```
RedFox20, line 369: Do we need to use decimal for this calculation? What's the accuracy difference compared to `pricePerYield / pricePerPrincipal` ?

test/TempusController.test.ts
```diff
@@ -314,6 +315,93 @@ describeForEachPool("TempusController", (testPool:PoolTestFixture) =>
...
 
+    it("Should successfully swap Yields --> Principals w/ 3% Maximum Slippage", async () => 
+    {
+      await initAMM(user1, /*ybtDeposit*/1000000, /*principals*/100000, /*yields*/1000000);
+      await testPool.controller.depositYieldBearing(owner, testPool.tempus, 100);
+
+      const pricePerYield = await testPool.yields.getPricePerFullShareStored();
+      const pricePerPrincipal = await testPool.principals.getPricePerFullShareStored();
+      
+      const currentYieldsRate = new Decimal(pricePerYield.toString()).div(pricePerPrincipal.toString()).toFixed(testPool.principals.decimals); /// TODO: move Decimal.js usage to a separate math helper
```
RedFox20, line 326: Yeah, this is duplicated in 3 places, so it should be a function one level above, before the `it("Should success...`
```ts
    async function currentYieldsRate(): Promise<NumberOrString> {
      const pricePerYield = await testPool.yields.getPricePerFullShareStored();
      const pricePerPrincipal = await testPool.principals.getPricePerFullShareStored();
      return pricePerYield / pricePerPrincipal;
    }
```

test/TempusController.test.ts
```diff
@@ -314,6 +315,93 @@ describeForEachPool("TempusController", (testPool:PoolTestFixture) =>
...
+      expect(await controller.exitTempusAmmAndRedeem(
+        testPool, 
+        owner, 
+        "0", 
+        "0",
+        await testPool.yields.balanceOf(owner),
+        false,
+        currentYieldsRate.toString(),
+        "0.03"
```
RedFox20, line 335: For `0` and `0.03` there is no need to use a string

contracts/TempusController.sol
```diff
@@ -667,31 +675,31 @@ contract TempusController is ReentrancyGuard, Ownable, Versioned {
...
+                (uint256 swapAmount, bool yieldsIn) = tempusAMM.getSwapAmountToEndWithEqualShares(
+                    principals,
+                    yields,
+                    maxLeftoverShares
+                );
+                uint256 minReturn = yieldsIn
+                    ? swapAmount.mulfV(yieldsRate, tempusPool.backingTokenONE())
+                    : swapAmount.divfV(yieldsRate, tempusPool.backingTokenONE());
+
+                minReturn = minReturn.mulfV(1e18 - maxSlippage, 1e18);
```
RedFox20, line 695: `(6 + 18) - 18 = 6` precision checks out 👍🏻

test/utils/TempusController.ts
```diff
@@ -255,7 +257,8 @@ export class TempusController extends ContractBase {
       0,
       0,
       maxLeftoverShares,
-      0,
+      amm.principalShare.toBigNum(yieldsRate),
+      parseDecimal(maxSlippage, 18),
```
RedFox20, line 261: Should use `toWei` here instead

contracts/TempusController.sol
```diff
@@ -283,19 +279,32 @@ contract TempusController is ReentrancyGuard, Ownable, Versioned {
         uint256 minPrincipalsStaked,
         uint256 minYieldsStaked,
         uint256 maxLeftoverShares,
-        uint256 minRate,
+        uint256 yieldsRate,
+        uint256 maxSlippage,
         bool toBackingToken,
         uint256 deadline
     ) external nonReentrant {
```
axic, line 286: As discussed I think we should take the opportunity to rename this function, as it is called `exitTempusAmmAndRedeem` and we also have `exitTempusAMMAndRedeem`.

I suggest `exitTempusAMMCompletelyAndRedeem` or `completelyExitTempusAMMAndRedeem`.

yuval-weiss: Do you think the other `exitTempusAMMAndRedeem` shouldn't be renamed?

test/TempusController.test.ts
yuval-weiss: I think we shouldn't worry about it tests formatting at all right now... I started working on a PR that adds eslint you our codebase so we just dictate some style rules and auto-enforce and fix all the styling issues we have...

test/TempusController.test.ts
yuval-weiss: I'm not sure what's the accuracy difference... I do think we should use Decimal.js whenever we can to prevent any precision issues. Imo we should add a math helper (`.sub()`, `.mul()` etc...) that uses Decimal.js

contracts/TempusController.sol
axic: I'd minimize the risks, the other one *seems* fine and the calling code of it is not changing in the frontend, as opposed to this one.

test/TempusController.test.ts
yuval-weiss: Done

contracts/TempusController.sol
mijovic: Why `completelyExitTempusAMMAndRedeem`?

test/TempusController.test.ts
```diff
@@ -314,6 +315,110 @@ describeForEachPool("TempusController", (testPool:PoolTestFixture) =>
...
+      await expect(controller.exitTempusAmmAndRedeem(
+        testPool, 
+        owner, 
+        0, 
+        0,
+        await testPool.yields.balanceOf(owner),
+        false,
+        await calculateCurrentYieldsRate(),
+        "0.001"
+      )).to.be.revertedWith("BAL#507");
```
RedFox20:
```ts
async function exitAmmAndRedeem(user:Signer, share:ERC20, maxSlippage:NumberOrString): Promise<Transaction> {
    const balance = await share.yields.balanceOf(user);
    const yieldsRate = await calculateCurrentYieldsRate();
    return controller.exitTempusAmmAndRedeem(testPool, user, 0, 0, balance, false, yieldsRate, maxSlippage);
}
```

```ts
      await expect(exitAmmAndRedeem(owner, testPool.yields, "0.001")).to.be.revertedWith("BAL#507");
```

test/TempusController.test.ts
```diff
@@ -314,6 +315,110 @@ describeForEachPool("TempusController", (testPool:PoolTestFixture) =>
...
+    });
+
+    it("Should fail swap due to minimum return Yields --> Principals w/ 0.1% Maximum Slippage", async () => 
+    {
+      await initAMM(user1, /*ybtDeposit*/1000000, /*principals*/100000, /*yields*/1000000);
+      await testPool.controller.depositYieldBearing(owner, testPool.tempus, 100);
+
+      await expect(controller.exitAmmGivenLpAndRedeem(
```
RedFox20, line 374:
```ts
async function exitAmmAndRedeem(user:Signer, share:ERC20, maxSlippage:NumberOrString): Promise<Transaction> {
    const balance = await share.yields.balanceOf(user);
    const yieldsRate = await calculateCurrentYieldsRate();
    return controller.exitTempusAmmAndRedeem(testPool, user, 0, 0, balance, false, yieldsRate, maxSlippage);
}
```
```ts
      await expect(exitAmmAndRedeem(owner, testPool.yields, "0.001")).to.be.revertedWith("BAL#507");
```

yuval-weiss: I don't think it makes much sense because some of the tests use different arguments for `exitTempusAmmAndRedeem` (yieldsRate different than the `CurrentYieldsRate`, more than 1 share balance as an argument, etc...)

mijovic:
Some unit tests are failing, otherwise looks good to go

PR #464, opened 2021-12-10
pool: Consistently use the type uint256 over uint

axic:
This ensures we use the types consistently, as currently we have a large mix.

Still, I have avoided touched the mocks, which is also using a mix.

### DIFF
--- a/contracts/TempusPool.sol
+++ b/contracts/TempusPool.sol
@@ -418,7 +418,7 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable, Versioned
     /// pricePerYield = currentYield * (estimatedYield - 1) / (estimatedYield)
     /// Return value decimal precision in backing token precision
     function pricePerYieldShare(uint256 currYield, uint256 estYield) private view returns (uint256) {
-        uint one = exchangeRateONE;
+        uint256 one = exchangeRateONE;
         // in case we have estimate for negative yield
         if (estYield < one) {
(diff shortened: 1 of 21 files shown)


RedFox20 (changes requested):
Why should we consistently use `uint256`, what's the benefit? It mostly just destroys formatting in specific parts of the code

axic:
> Why should we consistently use uint256, what's the benefit?

The benefit is consistency and less questions. Either we commit to `uint256` or to `uint`, as otherwise questions arise, such as "so is uint not 256-bits?".

> It mostly just destroys formatting in specific parts of the code

I knew you're going to get put off by those three instances :)

yuval-weiss (approved):

contracts/TempusPool.sol
```diff
@@ -520,10 +520,20 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable, Versioned
     /// @return Stored Interest Rate, decimal precision depends on specific TempusPool implementation
     function currentInterestRate() public view virtual override returns (uint256);
 
-    function numYieldTokensPerAsset(uint backingTokens, uint interestRate) public view virtual override returns (uint);
+    function numYieldTokensPerAsset(uint256 backingTokens, uint256 interestRate)
```
RedFox20, line 523: Figure out a way where it doesn't destroy formatting like here and then it'll be ok to merge

PR #467, opened 2021-12-10
controller: Ensure that untrustedTransferFrom is not called on backingToken(0)

axic:
Depends on #466.

### DIFF
--- a/test/TempusPool.DepositBackingTokens.test.ts
+++ b/test/TempusPool.DepositBackingTokens.test.ts
@@ -22,24 +22,37 @@ describeForEachPool("TempusPool DepositBackingTokens", (pool:PoolTestFixture) =>
 
   it("Should revert on Eth transfer as BT when not accepted", async () =>
   {
-    await pool.createDefault();
-    const [owner] = pool.signers;
-
     if (!pool.acceptsEther) {
+      await pool.createDefault();
+      const [owner] = pool.signers;
+
       (await pool.expectDepositBT(owner, 100, owner, /*ethValue*/ 1)).to.equal("given TempusPool's Backing Token is not ETH");
     }
   });
 
   it("Should revert on mismatched Eth transfer as BT when it is accepted", async () =>
   {
-    await pool.createDefault();
-    const [owner] = pool.signers;
-
     if (pool.acceptsEther) {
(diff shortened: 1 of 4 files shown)


test/TempusPool.DepositBackingTokens.test.ts
```diff
@@ -43,6 +43,19 @@ describeForEachPool("TempusPool DepositBackingTokens", (pool:PoolTestFixture) =>
     }
   });
 
+  it("Should revert on mismatched Eth transfer (with 0 value) as BT when it is accepted", async () =>
+  {
+    await pool.createDefault();
+    const [owner] = pool.signers;
```
RedFox20, line 49: I think since this test only works if `acceptsEther`, we shouldn't even create the contract instance to speed up other tests.
So, just moving `pool.createDefault()` inside the if block will do the trick

test/pool-utils/PoolTestFixture.ts
```diff
@@ -220,9 +220,10 @@ export abstract class PoolTestFixture {
   /**
    * Deposit BackingTokens into TempusPool
    */
-  async depositBT(user:Signer, backingTokenAmount:NumberOrString, recipient:SignerOrAddress = user, ethValue: NumberOrString = 0): Promise<Transaction> {
+  async depositBT(user:Signer, backingTokenAmount:NumberOrString, recipient:SignerOrAddress = user, ethValue: NumberOrString = undefined): Promise<Transaction> {
     // NOTE: this means tests can't use ethValue=0 forcefully to test special cases
-    const ethToTransfer = this.acceptsEther ? (ethValue || backingTokenAmount) : ethValue;
+    const ethToTransfer = this.acceptsEther ? ((ethValue == undefined) ? backingTokenAmount : ethValue) : (ethValue || 0);
```
RedFox20, line 225: I guess this works. What's the plan with the other calls that accept `ethValue` ?

mijovic (dismissed):

test/TempusPool.DepositBackingTokens.test.ts
axic: Done.

RedFox20 (approved):

PR #488, opened 2022-01-07
pools: Add safety check on decimal precision to Yearn

axic:


contracts/pools/YearnTempusPool.sol
```diff
@@ -40,6 +40,8 @@ contract YearnTempusPool is TempusPool {
             maxFeeSetup
         )
     {
+        require(vault.decimals() == IERC20Metadata(vault.token()).decimals(), "Decimals precision mismatch");
```
axic, line 43: Enforcing these because the interest rate calculation below assumes it.

contracts/protocols/yearn/IYearnVaultV2.sol
```diff
@@ -2,19 +2,21 @@
 pragma solidity 0.8.10;
 
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
 
 /***
     @notice based on https://github.com/yearn/yearn-vaults/blob/main/contracts/Vault.vy
  */
-interface IYearnVaultV2 is IERC20 {
+interface IYearnVaultV2 is IERC20, IERC20Metadata {
```
axic, line 10: Proof: https://github.com/yearn/yearn-vaults/blob/main/contracts/Vault.vy#L66-L68

yuval-weiss: Can we remove IERC20 and only inherit from IERC20Metadata?

contracts/mocks/yearn/YearnVaultMock.sol
```diff
@@ -49,7 +49,7 @@ contract YearnVaultMock is ERC20, IYearnVaultV2 {
     }
 
     // yTokens always have the same decimals as their corresponding underlying tokens
-    function decimals() public view override returns (uint8) {
+    function decimals() public view override(ERC20, IERC20Metadata) returns (uint8) {
```
yuval-weiss, line 52: Can probably be removed if IYearnVault only inherits `IERC20Metadata`

contracts/protocols/yearn/IYearnVaultV2.sol
axic: Just tried but it doesn't actually help with https://github.com/tempus-finance/tempus-protocol/pull/488#discussion_r782864710, so I'd just keep it.

yuval-weiss: I think it makes sense to remove it any any case, no? `IERC20Metadata` is a parent of `IERC20`

contracts/mocks/yearn/YearnVaultMock.sol
yuval-weiss: Closing this.

contracts/pools/YearnTempusPool.sol
yuval-weiss: Noice

contracts/protocols/yearn/IYearnVaultV2.sol
axic: We use the same in other places and since it is an extension (in name) I think it makes sense to keep it. Would reconsider if we get issues down the line.

yuval-weiss: Alright ser, sounds good to me.

yuval-weiss (approved):

PR #490, opened 2022-01-10
pools: Manually check actual yieldBearingToken balance upon Yearn deposit

axic:


contracts/pools/YearnTempusPool.sol
```diff
@@ -52,7 +52,11 @@ contract YearnTempusPool is TempusPool {
         // Deposit to Yearn Vault
         IERC20(backingToken).safeIncreaseAllowance(address(yearnVault), amount);
 
-        return yearnVault.deposit(amount);
+        uint256 preDepositBalance = IERC20(yieldBearingToken).balanceOf(address(this));
+        yearnVault.deposit(amount);
```
axic, line 56: Should there be an assertion here that this can't return 0?

contracts/pools/YearnTempusPool.sol
```diff
@@ -52,7 +52,11 @@ contract YearnTempusPool is TempusPool {
...
+        uint256 postDepositBalance = IERC20(yieldBearingToken).balanceOf(address(this));
```
yuval-weiss, line 57: This manual check should also be done in `withdrawFromUnderlyingProtocol`

axic: Actually I don't think so, because that directly retrives it to the recipient.

yuval-weiss: In the Yearn implementation, you mean? If so, I think it's still a good idea to do this calculation ourselves... Their implementation might change in the future so better to just do it on our end to make sure it's bulletproof.

axic: No, I mean the withdrawal code takes a `recipient`. We do not check balances in other pools if the recipient is not the pool.

yuval-weiss: Ok, how about checking the `recipient`'s balance before and after, and then `assert`ing it's equal to the number returned from Yearn?

yuval-weiss (approved):

PR #492, opened 2022-01-17
pool: Optimizing amount of updateInterestRate calls

mijovic:


contracts/amm/TempusAMM.sol
```diff
@@ -651,7 +651,8 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
     function _getTokenRates() private returns (uint256[] memory) {
         uint256[] memory rates = new uint256[](_TOTAL_TOKENS);
         rates[0] = _token0.getPricePerFullShare();
-        rates[1] = _token1.getPricePerFullShare();
+        // We already did updateInterestRate, so we can use stored values
```
RedFox20, line 654: Nice! 🧠

contracts/TempusPool.sol
```diff
@@ -422,15 +418,17 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable, Versioned
     }
 
     function pricePerYieldShare() external override returns (uint256) {
-        return pricePerYieldShare(currentYield(), estimatedYield());
+        uint256 yield = currentYield();
+        return pricePerYieldShare(yield, estimatedYield(yield));
     }
 
     function pricePerYieldShareStored() external view override returns (uint256) {
         return pricePerYieldShare(currentYieldStored(), estimatedYieldStored());
```
RedFox20, line 426: Maybe update this as well so the code looks similar to your updates? Less calls should reduce gas a bit as well.
```ts
uint256 yield = currentYieldStored();
return pricePerYieldShare(yield, estimatedYield(yield));
```

contracts/amm/TempusAMM.sol
axic: From a clarity perspective, having `pool.updateInterest()` and then using `_token0.getPricePerFullShare(); _token1.getPricePerFullShare();` would be cool.

RedFox20: Expanding on your idea, the most optimal way to get those rates would be to have:
`tempusPool.getTokenPricesPerFullShare()` and `tempusPool.getTokenPricesPerFullShareStored()`

contracts/TempusPool.sol
mijovic: We can do this in separate PR

contracts/amm/TempusAMM.sol
mijovic: This can be done in separate PR

PR #502, opened 2022-01-21
principal rate orcale

RedFox20:


contracts/math/ExponentialMovingAverage.sol
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/// Implementation for Exponential Weighted Moving Average
+/// The weight must be calculated from updateInterval and turnoverTime
+library ExponentialMovingAverage {
+    /// @dev Creates a new EMA State
+    /// @param updateInterval time between updates
+    /// @param newValueWeight weight for new values
+    function create(uint256 updateInterval, uint256 newValueWeight) internal returns (EMAState memory) {
+        return EMAState(0, 0, updateInterval, newValueWeight);
```
axic, line 19: Perhaps add something like this:
```suggestion
        require(newValueWeight < 1e18, "Invalid weight.");
        return EMAState(0, 0, updateInterval, newValueWeight);
```

Or use assert for both this and interval to avoid programming errors?

contracts/amm/TempusAMM.sol
```diff
@@ -152,6 +162,12 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
         }
     }
 
+    /// @return Current Rate for Principals, averaged over time
+    function principalRateOracle() internal returns (uint256) {
```
axic, line 166: Why is this on the AMM and not on the Pool or Controller?

contracts/math/ExponentialMovingAverage.sol
RedFox20: Sure I can add some sanity checks here

contracts/amm/TempusAMM.sol
RedFox20: The simple reason is to keep this private, instead of exposing as public member of TempusPool/Controller. Would need long term planning to figure out where it should end up at

mijovic: We need to call this from private functions from amm so it updates actual oracle when needed.

contracts/amm/TempusAMM.sol
```diff
@@ -81,6 +83,8 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
 
     ITempusPool public immutable tempusPool;
 
+    EMAState private principalRateEMA;
```
yuval-weiss, line 86: Shouldn't this be public (or exposed via some public function)?

contracts/amm/TempusAMM.sol
```diff
@@ -152,6 +156,12 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
         }
     }
 
+    /// @return Current Rate for Principals, averaged over time
+    function principalRateOracle() internal returns (uint256) {
```
yuval-weiss, line 160: imo better to rename the function to something that includes the word `update`...

contracts/amm/TempusAMM.sol
```diff
@@ -152,6 +156,12 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
+        uint256 rate = tempusPool.pricePerPrincipalShareStored();
+        return ExpMovingAverage.update(principalRateEMA, rate, 2 * 60, 1e18 / 10);
```
yuval-weiss, line 162: `updateInterval` and `newValueWeight ` should probably be configurable in the constructor (we can also discuss exposing an admin function to modify these)

contracts/amm/TempusAMM.sol
RedFox20: We haven't really decided that yet, but it would have to be via some extra function

contracts/amm/TempusAMM.sol
RedFox20: `updatePrincipaRateOracle()` ?

contracts/amm/TempusAMM.sol
RedFox20: I was thinking about this, after we figure out these parameters, most likely it's going to stay the same everywhere, which is why I moved the arguments here. Alternatively I'd make them into constants.

This isn't something one would tweak on a whim to be honest

contracts/amm/TempusAMM.sol
mijovic, line 86:
```suggestion
    EMAState private principalRateEMAState;
```

contracts/amm/TempusAMM.sol
```diff
@@ -139,6 +143,8 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
         (_token0, _token1) = (token0, token1);
 
         tempusPool = pool;
+        // TODO: set updateInterval and weight from constructor
+        principalRateEMA.newValueWeight = 1e18 / 100;
```
mijovic, line 147: Maybe?
```suggestion
        principalRateEMA.newValueWeight = 1e16;
```

contracts/amm/TempusAMM.sol
```diff
@@ -152,6 +158,12 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
         }
     }
 
+    /// @return Current Rate for Principals, averaged over time
+    function updatePrincipalRateOracle() internal returns (uint256) {
+        uint256 rate = tempusPool.pricePerPrincipalShareStored();
```
mijovic, line 163: we need to properly calculate rate of principal here (this is just getting the weight from TempusPool).

We need to calculate it using balances (from Vault), amplification, and tokenRates, the same way like when we do a swap

contracts/math/MovingAverage.sol
```diff
@@ -0,0 +1,114 @@
...
+        uint256 newValue,
+        uint256 updateInterval
+    ) internal returns (uint256) {
+        uint256 ts = block.timestamp;
+        uint256 lastUpdate = ma.lastUpdateTime;
+        uint256 elapsed = ts - lastUpdate;
+        if (elapsed >= updateInterval) {
+            if (lastUpdate != 0) {
+                uint256 oldValue = ma.value;
+                uint256 n = elapsed / updateInterval;
```
mijovic, line 45: we need to add `require(updateInterval != 0)`

contracts/math/MovingAverage.sol
```diff
@@ -0,0 +1,114 @@
...
+                uint256 w = getBiasedWeight(oldValue, newValue, ma.newValueWeight, n);
+                // standard EMA
+                newValue = ((newValue * w) + (oldValue * (1e18 - w))) / 1e18;
```
mijovic, line 48: I hope this can't overflow. Otherwise I would use safe math library

contracts/math/MovingAverage.sol
```diff
@@ -0,0 +1,114 @@
...
+
+    function getBiasedWeight(
+        uint256 oldValue,
+        uint256 newValue,
+        uint256 newValueWeight,
+        uint256 iterations
+    ) private pure returns (uint256) {
+        uint256 bias;
+        uint256 distance = newValue >= oldValue ? newValue - oldValue : oldValue - newValue;
+        if (distance > (1e18 / 10)) {
```
mijovic, line 98: Are we assuming 18 decimals here?

contracts/math/MovingAverage.sol
RedFox20: Is it needed here? (would increase gas). Maybe `assert()` instead, because that would not take extra gas?

contracts/math/MovingAverage.sol
RedFox20: This strongly relies on values being around 1e18, so 1e18*1e18 + 1e18*1e18 will not overflow.

Currently this biased EMA only works around RATE values such as 0.9, 1.5, 2.5, ... not intended for larger numbers

contracts/math/MovingAverage.sol
RedFox20: Yes, all of this assumes 18 decimals

contracts/amm/TempusAMM.sol
RedFox20: Is there a cleaner way to represent `0.01` ?

contracts/math/MovingAverage.sol
axic: `assert` takes as much runtime gas as `require`, on the happy path. The only difference is in code size and on the failure path.

RedFox20: I added `require` then

PR #507, opened 2022-01-25
Pass Chainlink address instead of ENS to `totalValueLockedAtGivenRate`

danieledamiani:
First Solidity PR so please be gentle 😊


mijovic:
To me, it looks good. Let's wait for @yuval-weiss to do one more review

contracts/stats/ChainlinkTokenPairPriceFeed/ChainlinkTokenPairPriceFeed.sol
```diff
@@ -14,14 +14,13 @@ abstract contract ChainlinkTokenPairPriceFeed is ITokenPairPriceFeed {
...
-    function getRate(bytes32 chainlinkAggregatorNodeHash)
+    function getRate(address chainlinkAggregatorNode)
         public
         view
         override
         returns (uint256 rate, uint256 rateDenominator)
     {
-        IENSResolver ensResolver = ENS.resolver(chainlinkAggregatorNodeHash);
-        IChainlinkAggregator chainLinkAggregator = IChainlinkAggregator(ensResolver.addr(chainlinkAggregatorNodeHash));
+        IChainlinkAggregator chainLinkAggregator = IChainlinkAggregator(chainlinkAggregatorNode);
```
yuval-weiss, line 23: Better to pass the `chainLinkAggregatorProxy` (which holds a reference to the active aggregator) rather than the actual aggregator... It'll save you a call from the frontend (`proxy.aggregator()`) by querying the aggregator from the proxy on-chain.

https://docs.chain.link/docs/fantom-price-feeds/#Fantom%20Mainnet

contracts/stats/ChainlinkTokenPairPriceFeed/ChainlinkTokenPairPriceFeed.sol
```diff
@@ -14,14 +14,13 @@ abstract contract ChainlinkTokenPairPriceFeed is ITokenPairPriceFeed {
     // on other networks with a different ENS address.
     IENS private constant ENS = IENS(0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e);
```
yuval-weiss, line 15: Can be removed (inc. docs above)

yuval-weiss: And including IENS interface file.

contracts/stats/ChainlinkTokenPairPriceFeed/ChainlinkTokenPairPriceFeed.sol
yuval-weiss: after giving it another look I've noticed that this is alright because the proxy itself also exposes `latestRoundData`.

contracts/stats/ITokenPairPriceFeed.sol
```diff
@@ -3,8 +3,8 @@ pragma solidity 0.8.10;
 
 interface ITokenPairPriceFeed {
     /// Fetches the rate between a given token pair
-    /// @param rateConversionData Data that specifies the target tokens (each ITokenPairPriceFeed might have different input requirements)
+    /// @param chainlinkAggregatorNode Chainlink token pair address (each ITokenPairPriceFeed might have different input requirements)
```
yuval-weiss, line 6: This interface was meant to be generic.
I now think we can just get rid of this interface since we're not gonna have other types of price feeds (we'll probably migrate use The Graph and deprecate this completely....

yuval-weiss (approved):

mijovic (dismissed):

mijovic:
Please squash commits and we can merge this

yuval-weiss (approved):

PR #509, opened 2022-02-01
TempusVault prototype (for discussion)

axic:


contracts/vault/TempusVault.sol
```diff
@@ -0,0 +1,93 @@
...
+        require(vault.strategy() != address(this)); // TODO error code/message
+
+        // TODO: implement withdrawal of all assets here
+        assert(false);
+    }
+}
+
+/// A Tempus Vault is tied to a single yield bearing token, but supports
+/// pluggable strategies around those.
+contract TempusVault is Ownable, Versioned {
```
yuval-weiss, line 56: The vault would inherit from `ERC20` too I guess?

contracts/vault/TempusVault.sol
```diff
@@ -0,0 +1,93 @@
...
+
+    /// Replace the current strategy with a new modul. Checks version conformance.
+    function replaceStrategy(TempusVaultStrategy _strategy) public onlyOwner {
+        // Check compatibility.
+        Version memory version = _strategy.version();
+        // TODO create a helper in Versioned for this
+        //require(_strategy.major == _major && _strategy.minor >= _minor); // TODO error codes / messages
+
+        // Replace approvals.
+        yieldToken.safeApprove(address(strategy), 0);
```
yuval-weiss, line 77: IMO it's better to hold the funds in the strategy contract.

axic:
This replaceable strategy design will not go ahead now.

PR #511, opened 2022-02-03
pool: combined price per shares

RedFox20:


contracts/TempusPool.sol
```diff
@@ -82,7 +82,7 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable, Versioned
         TokenData memory principalsData,
         TokenData memory yieldsData,
         FeesConfig memory maxFeeSetup
-    ) Versioned(1, 0, 0) {
+    ) Versioned(1, 1, 0) {
```
axic, line 85:
```suggestion
    ) Versioned(2, 0, 0) {
```

I'd argue this is a breaking change as the functions are removed. If you'd keep the old functions too, just as wrappers over the new ones, then it wouldn't be breaking but an API extension.

contracts/TempusPool.sol
```diff
@@ -521,7 +495,7 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable, Versioned
     /// @dev This updates the underlying pool's interest rate
     ///      It should be done first thing before deposit/redeem to avoid arbitrage
     /// @return Updated current Interest Rate, decimal precision depends on specific TempusPool implementation
-    function updateInterestRate() internal virtual returns (uint256);
+    function updateInterestRate() public virtual returns (uint256);
```
axic, line 498: Is this only changed because of the need for Rari testing?

RedFox20: It is required for testing

contracts/TempusPool.sol
axic: Screenshot

If we follow semver, then the quesiton is super simple: does it break the API? If so, major bump.

contracts/TempusPool.sol
axic: I think if we are making this public, then it should become part of `ITempusPool` too. I think it is safe to expose this, as it just means whoever is paying the gas for it will do the job for the pool 😓

contracts/TempusPool.sol
```diff
@@ -423,47 +411,33 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable, Versioned
...
+    function getPricePerShare(uint256 interestRate) private view returns (uint256 principalsRate, uint256 yieldsRate) {
+        uint256 curYield = currentYield(interestRate);
+        uint256 estYield = estimatedYield(curYield);
         // in case we have estimate for negative yield
         if (estYield < exchangeRateONE) {
-            return interestRateToSharePrice(currYield);
+            principalsRate = interestRateToSharePrice(curYield);
+            yieldsRate = 0;
+        } else {
+            uint256 one = exchangeRateONE;
```
axic, line 427: I'd move this outside fo the branch, so that the immutable copy is done once in this function.

contracts/TempusPool.sol
```diff
@@ -423,47 +411,33 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable, Versioned
...
+    /// @return principalsRate Principals to BackingToken rate expressed in Backing Token decimal precision
+    /// @return yieldsRate Yields to BackingToken rate expressed in Backing Token decimal precision
+    function getPricePerShare(uint256 interestRate) private view returns (uint256 principalsRate, uint256 yieldsRate) {
+        uint256 curYield = currentYield(interestRate);
+        uint256 estYield = estimatedYield(curYield);
         // in case we have estimate for negative yield
         if (estYield < exchangeRateONE) {
-            return interestRateToSharePrice(currYield);
+            principalsRate = interestRateToSharePrice(curYield);
+            yieldsRate = 0;
```
axic, line 425: `yieldsRate` (and other return parameters) is pre-initialized to zero already, this line may not be deduplicated by the compiler

contracts/TempusPool.sol
```diff
@@ -391,18 +391,6 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable, Versioned
...
-        return currentYield(updateInterestRate());
-    }
-
-    function currentYieldStored() private view returns (uint256) {
-        return currentYield(currentInterestRate());
-    }
-
-    function estimatedYieldStored() private view returns (uint256) {
-        return estimatedYield(currentYieldStored());
-    }
```
mijovic, line 404: This was probably easier to remove in separate PR, but let's leave it here

contracts/TempusPool.sol
```diff
@@ -423,47 +411,32 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable, Versioned
...
 
-    /// pricePerYield = currentYield * (estimatedYield - 1) / (estimatedYield)
-    /// Return value decimal precision in backing token precision
-    function pricePerYieldShare(uint256 currYield, uint256 estYield) private view returns (uint256) {
+    /// @dev Calculates Principals and Yields exchange rate
+    /// principalsRate = currentYield / estimatedYield
+    /// yieldsRate = currentYield * (estimatedYield - 1) / (estimatedYield)
+    /// @return principalsRate Principals to BackingToken rate expressed in Backing Token decimal precision
+    /// @return yieldsRate Yields to BackingToken rate expressed in Backing Token decimal precision
+    function getPricePerShare(uint256 interestRate) private view returns (uint256 principalsRate, uint256 yieldsRate) {
```
mijovic, line 419: Maybe?
```suggestion
    function getPricePerShares(uint256 interestRate) private view returns (uint256 principalsRate, uint256 yieldsRate) {
```

contracts/amm/TempusAMM.sol
```diff
@@ -62,6 +62,8 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
 
     IPoolShare internal immutable _token0;
     IPoolShare internal immutable _token1;
+    IPoolShare internal immutable _principals;
+    IPoolShare internal immutable _yields;
```
mijovic, line 66: I would not change this here. Would try to stick with `_token0` and `_token1` as it was for now

PR #512, opened 2022-02-03
Introduce solidity-docgen

axic:
This is based on the docgen rewrite: https://github.com/OpenZeppelin/solidity-docgen/tree/rewrite

Run it with `npx hardhat docgen`.


axic:
This should be easier now as it is a single Solidity version after #591. That was one of the reasons this was semi-blocked.

PR #518, opened 2022-02-10
Add deployment JSON configuration

axic:


scripts/pool_FYEDA002.json
```diff
@@ -0,0 +1,48 @@
...
+  "principal": {
+    "name": "Tempus Capital Token-FYEDA002",
+    "symbol": "tCapital-FYEDA002"
+  },
+  "yield": {
+    "name": "Token Tempus Yield Token-FYEDA002",
+    "symbol": "tYield-FYEDA002"
+  },
+  "pool": {
+    "maturity": {
```
axic, line 22: I think we could use an ISO date here and parse it.

scripts/pool_FYEDA002.json
```diff
@@ -0,0 +1,48 @@
...
+      "year": 2022,
+      "month": 8,
+      "day": 26,
+      "hour": 8,
+      "minute": 0
+    },
+    "estimatedYield": "0.0347",
```
axic, line 29: Need to decide whether to rely on Javscript floats or parsed strings.

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,146 @@
...
+    poolBacking,
+    tempusController,
+    maturityTimestamp,
+    ethers.utils.parseUnits(estimatedYield, exchangeRatePrecision).toString(),
+    /*principalsData*/{
+      name: config.principal.name,
+      symbol: config.principal.symbol
+    },
+    /*yieldsData*/{
+      name: config["yield"].name,
```
axic, line 83: `yield` is a Javascript keyword

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,146 @@
...
+
+const MONTH = 60 * 60 * 24 * 30;
+
+async function deploy(configName) {
+  const config = JSON.parse(readFileSync(configName, 'utf-8').toString());
+
+  console.log(config);
+
+  const deployerPrivateKey = "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff";
+//  const deployerPrivateKey = await utils.promptPrivateKey("Enter deployer Private Key");
```
yuval-weiss, line 14: uncomment?

scripts/pool_FYEDA002.json
yuval-weiss: yeah I think it's a good idea

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,146 @@
...
+  var poolBacking;
+  var customPoolConstructorArgs = [];
+  if (config.lido) {
+    contractName = "LidoTempusPool";
+    poolBacking = config.lido.steth;
+    customPoolConstructorArgs = [ config.lido.referrer ];
+  } else if (config.yearn) {
+    contractName = "YearnTempusPool";
+    poolBacking = config.yearn.vault;
+  } else if (config.rari) {
```
yuval-weiss, line 55: Rari has a "custom" constructor arg `address backingToken` as a second parameter (:hidethepainharold:) so this would fail. We should change the order of all the pools' constructors so that extra/custom protocol-specific constructor arguments are appended last (and then this logic would work).

In the meantime we can also try to find a workaround.

scripts/pool_FYEDA002.json
yuval-weiss: Strings imo. We can get some weird behaviors with javascript floats.

scripts/deploy.pool.config.ts
axic: Oh no, you're correct. Rari is always an exception 😬

scripts/deploy.pool.config.ts
axic: Yes, need to do that before merging :)

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,146 @@
...
+  const config = JSON.parse(readFileSync(configName, 'utf-8').toString());
+
+  console.log(config);
+
+  const deployerPrivateKey = "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff";
+//  const deployerPrivateKey = await utils.promptPrivateKey("Enter deployer Private Key");
+
+  const tempusController = config.controller;
+
+  const maturityYear = config.pool.maturity.year;
```
RedFox20, line 18: Is there a reason to pull out these vars? Why not just use `config.pool.maturity.year` etc. directly?

scripts/pool_FYEDA002.json
RedFox20: Definitely 💯

scripts/pool_FYEDA002.json
RedFox20: Strings. And then we can properly parse the decimal using correct token precision. We should connect to the underlying tokens and validate it of course

scripts/pool_FYEDA002.json
```diff
@@ -0,0 +1,48 @@
+{
+  "yearn": {
+    "vault": "0x637eC617c86D24E421328e6CAEa1d92114892439"
+  },
+  "decimals": 18,
+  "ybt": {
+    "decimals": 18,
```
RedFox20, line 7: Is there a way to connect to the contract and validate this by fetching decimals from the chain?

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,146 @@
...
+    /*principalsData*/{
+      name: config.principal.name,
+      symbol: config.principal.symbol
+    },
+    /*yieldsData*/{
+      name: config["yield"].name,
+      symbol: config["yield"].symbol
+    },
+    /*maxFeeSetup:*/{
+      depositPercent:      ethers.utils.parseUnits(depositMaxFee, ybtPrecision).toString(),
```
RedFox20, line 87: We should not use `ethers` directly. Lets use our wrapper utils like `parseDecimal()` from `Decimal.ts`.

There is reason for that - it will make it easier to upgrade our Decimals library in the near future 🙂

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,146 @@
...
+  const amplificationStart = config.amm.initialAmplificationFactor;
+  const amplificationEnd = config.amm.finalAmplificationFactor;
+  const swapFeePercentage = config.amm.swapFee;
+  const pauseWindowDuration = config.amm.pauseWindowDuration;
+  const bufferPeriodDuration = config.amm.bufferPeriodDuration;
+  const owner = config.amm.owner;
+
+  const exchangeRatePrecision = config.decimals;
+
+  const ybtPrecision = config.ybt.decimals;
```
RedFox20, line 41: YBT decimals needs to be validated by fetching `decimals()` from the chain 👍🏻

scripts/deploy.pool.config.ts
axic: Pushed a workaround.

scripts/pool_FYEDA002.json
```diff
@@ -0,0 +1,42 @@
...
+  "principal": {
+    "name": "Tempus Capital Token-FYEDA002",
+    "symbol": "tCapital-FYEDA002"
+  },
+  "yield": {
+    "name": "Token Tempus Yield Token-FYEDA002",
+    "symbol": "tYield-FYEDA002"
+  },
+  "pool": {
+    "maturity": "2022-08-26T08:00:00.000+00:00",
```
axic, line 22: The annoying part with using `Date.parse` is that it needs either `yyyy-mm-dd` or this long format to enforce UTC, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#date_time_string_format

scripts/pool_FYEDA002.json
```diff
@@ -0,0 +1,42 @@
+{
+  "yearn": {
+    "vault": "0x637eC617c86D24E421328e6CAEa1d92114892439"
+  },
+  "decimals": 18,
+  "ybt": {
```
yuval-weiss, line 6: should we maybe replace the entire object with the ybt address, and then query all this data from the contract?

axic: Potentially, but what @RedFox20 suggest may be better: have it here, but also check it against on chain data. That way we do not lose visibility.

scripts/pool_FYEDA002.json
yuval-weiss: the long formatting is not too bad imo

scripts/pool_FYEDA002.json
yuval-weiss: related to my comment above - instead of validating we can just provide the address and fetch all the data we need

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  const depositMaxFee = parseDecimal(validateRange(config.pool.fees.deposit, 0, 0.05).toString(), ybtPrecision).toString();
+  const earlyRedemptionMaxFee = parseDecimal(validateRange(config.pool.fees.earlyRedemption, 0, 0.05).toString(), ybtPrecision).toString();
+  const matureRedemptionMaxFee = parseDecimal(validateRange(config.pool.fees.maturedRedemption, 0, 0.05).toString(), ybtPrecision).toString();
+
+  const balancerVault = config.vault;
+
+  const lpName = config.amm.lp.name;
+  const lpSymbol = config.amm.lp.symbol;
+  const amplificationStart = validateRange(config.amm.initialAmplificationFactor, 0, 1000);
```
RedFox20, line 44: I think there's no need to return the validate value, just `validateRange(config.amm.initialAmplificationFactor, 0, 1000);` which will terminate on failure. And later use `config.amm.initialAmplificationFactor`

axic: Sure, I had it first because using these conversions in-line without having local variables. If we keep the variables and/or keep floatong point, it works well.

scripts/pool_FYEDA002.json
RedFox20: The upside of duplication here is that we can validate that our assumptions are correct. If we assign an incorrect token address, it will fetch that invalid data and simply use it.

If we put decimals, name, symbol and validate all of them, then we can't make such a fatal mistake

scripts/pool_FYEDA002.json
RedFox20: Also, I think according to documentation, this will work as well: `".000Z"`

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,150 @@
...
+  const amplificationEnd = validateRange(config.amm.finalAmplificationFactor, 0, 1000);
+  const swapFeePercentage = parseDecimal(validateRange(config.amm.swapFee, 0, 0.03).toString(), 18).toString();
```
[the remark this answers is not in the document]
RedFox20: It is not needed, ethers will do it automatically. BigNumber is actually a string anyway

scripts/pool_FYEDA002.json
axic: This is handled now.

scripts/deploy.pool.config.ts
axic: This is handled now.

scripts/deploy.pool.config.ts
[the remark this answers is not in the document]
axic: Removed.

test/utils/ERC20.ts
```diff
@@ -41,13 +41,36 @@ export class ERC20 extends ContractBase implements IERC20 {
...
    * Attaches to any contract address and attempts to convert it to ERC20
+   * Uses default provider
    * @param contractName Name of the solidity contract
    * @param contractAddress Address of the contract
    * @param decimals Contract decimals
    */
   static async attach(contractName:string, contractAddress:string, decimals:number): Promise<ERC20> {
     const contract = await this.attachContract(contractName, contractAddress);
-    return new ERC20(contractName, decimals, contract);
+    let erc20 = new ERC20(contractName, decimals, contract);
```
yuval-weiss, line 51:
```suggestion
    const erc20 = new ERC20(contractName, decimals, contract);
```

test/utils/ERC20.ts
```diff
@@ -41,13 +41,36 @@ export class ERC20 extends ContractBase implements IERC20 {
...
+   * @param contractName Name of the solidity contract
+   * @param contractAddress Address of the contract
+   * @param signer Signer to attach with, can be ethers.VoidSigner or SignerWithAddress
+   */
+  static async attachWithSigner(contractName:string, contractAddress:string, signer:AbstractSigner): Promise<ERC20> {
+    if (!signer) {
+      throw new Error("attachWithSigner: `signer` must not be null");
+    }
+    const contract = await this.attachContractWithSigner(contractName, contractAddress, signer);
+    let erc20 = new ERC20(contractName, /*decimals*/0, contract);
```
yuval-weiss, line 70:
```suggestion
    const erc20 = new ERC20(contractName, /*decimals*/0, contract);
```

test/amm/TempusAMM.test.ts
```diff
@@ -84,6 +84,10 @@ describeForEachPool("TempusAMM", (testFixture:PoolTestFixture) =>
     expect(+tokenIn.fromBigNum(postSwapTokenOutBalance.sub(preSwapTokenOutBalance))).to.be.within(+swapTest.swapAmountOut * 0.97, +swapTest.swapAmountOut * 1.02);
   }
 
+  it("[amplification] Fail on invalid inputs", async () => {
```
yuval-weiss, line 87: should this be here?

test/utils/ContractBase.ts
```diff
@@ -1,8 +1,10 @@
 import { ethers } from "hardhat";
-import { Contract, BigNumber } from "ethers";
-import { NumberOrString, parseDecimal, formatDecimal, MAX_NUMBER_DIGITS, DecimalConvertible } from "./Decimal";
+import { Contract } from "ethers";
+import { DecimalConvertible } from "./Decimal";
 import * as signers from "@nomiclabs/hardhat-ethers/dist/src/signers";
+import * as abstract_signer from "@ethersproject/abstract-signer/src.ts";
```
yuval-weiss, line 5:
```suggestion
import * as abstractSigner from "@ethersproject/abstract-signer/src.ts";
```

package.json
```diff
@@ -63,6 +63,7 @@
     "deposit-local:fork:mainnet": "cross-env HARDHAT_FORK=mainnet npx hardhat run scripts/deposit.local.fork.ts --network localhost",
     "deploy-ci:fork:mainnet": "cross-env HARDHAT_FORK=mainnet npx hardhat run scripts/deploy.local.fork.ts",
     "deploy-local": "npx hardhat run scripts/deploy.local.ts --network localhost",
+    "deploy-pool:mainnet": "cross-env ETH_NODE_URI_MAINNET=https://eth-mainnet.alchemyapi.io/v2/yInmJCdHvCzXe4qAxfPMDFgrbYIGsUhJ npx hardhat run scripts/deploy.pool.config.ts --network mainnet",
     "deploy-ci": "npx hardhat run scripts/deploy.local.ts",
     "deploy:controller": "npx hardhat run scripts/deploy.controller.ts",
     "deploy:stats": "npx hardhat run scripts/deploy.stats.ts",
```
yuval-weiss, line 69: `deploy:pool:lido` can be removed

yuval-weiss: (including the javascript script file)

test/amm/TempusAMM.test.ts
axic: No, it is a stray thing. I wanted to create tests for those.

yuval-weiss:
idk if you tested it on all of the different pools, but I think the best way to test this would be to start local mainnet fork hardhat node and then deploy different pools with `--network localhost`

package.json
```diff
@@ -63,6 +63,7 @@
     "deposit-local:fork:mainnet": "cross-env HARDHAT_FORK=mainnet npx hardhat run scripts/deposit.local.fork.ts --network localhost",
     "deploy-ci:fork:mainnet": "cross-env HARDHAT_FORK=mainnet npx hardhat run scripts/deploy.local.fork.ts",
     "deploy-local": "npx hardhat run scripts/deploy.local.ts --network localhost",
+    "deploy-pool:mainnet": "cross-env ETH_NODE_URI_MAINNET=https://eth-mainnet.alchemyapi.io/v2/yInmJCdHvCzXe4qAxfPMDFgrbYIGsUhJ npx hardhat run scripts/deploy.pool.config.ts --network mainnet",
```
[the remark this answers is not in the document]
axic: Not sure, this part was added by @RedFox20

scripts/pool_FYEDA002.json
```diff
@@ -0,0 +1,41 @@
...
+  "kind": "Yearn",
+  "decimals": 18,
+  "ybt": {
+    "address": "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
+    "decimals": 18,
+    "name": "DAI yVault",
+    "symbol": "yvDAI"
+  },
+  "controller": "0x8c47924b35C3667F59Df579F3ec061F8d7603242",
+  "vault": "0x637eC617c86D24E421328e6CAEa1d92114892439",
```
yuval-weiss, line 11: can we rename this to `ammVault`? `vault` can be easily confused with the Yearn Vault....

yuval-weiss: actually, better to just move `vault` into the `amm` object below.

package.json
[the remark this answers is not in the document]
RedFox20: This is just a placeholder, it should be removed 😬

scripts/pool_FYEDA002.json
axic: Yes, moving under amm may be nicer. Done.

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,245 @@
...
+  return config;
+}
+
+async function deploy(configName:string) {
+  let config:Config = JSON.parse(readFileSync(configName, 'utf-8').toString());
+
+  console.log(config);
+
+  const deployerPrivateKey = "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff";
+//  const deployerPrivateKey = await utils.promptPrivateKey("Enter deployer Private Key");
```
axic, line 232: This needs to be flipped before merging.

scripts/pool_Lido31March2022.json
```diff
@@ -0,0 +1,44 @@
...
+    "maturity": "2022-03-31T16:00:00.000+00:00",
+    "estimatedYield": "0.0347",
+    "fees": {
+      "deposit": 0.00,
+      "earlyRedemption": 0.00,
+      "maturedRedemption": 0.00
+    }
+  },
+
+  "amm": {
```
axic, line 31: Haven't filled out the amm part properly yet.

test/utils/ERC20.ts
```diff
@@ -1,6 +1,6 @@
 import { Contract } from "ethers";
 import { NumberOrString } from "./Decimal";
-import { ContractBase, SignerOrAddress, Signer, addressOf } from "./ContractBase";
+import { ContractBase, SignerOrAddress, Signer, AbstractSigner, addressOf } from "./ContractBase";
```
RedFox20, line 3: This file should be all good, simple changes ✅

test/utils/ContractBase.ts
```diff
@@ -1,8 +1,10 @@
 import { ethers } from "hardhat";
-import { Contract, BigNumber } from "ethers";
```
RedFox20, line 2: This file should be all good, simple changes ✅

scripts/pool_FYEDA002.json
RedFox20: Did you try with ".000Z" ? @axic

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,245 @@
...
+
+interface LidoConfig {
+  steth: string;
+}
+
+interface RariConfig {
+  fundManager: string;
+}
+
+interface Config {
```
RedFox20, line 55: This is really nice. 💯  I do believe as the next logical step we should move it to another file to reduce clutter in this file 🙂

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,245 @@
...
+    process.exit(1);
+  }
+  if ((await ybt.name()) !== ybtName) {
+    console.log(chalk.red("YBT: name mismatch"));
+    process.exit(1);
+  }
+  if ((await ybt.symbol()) !== ybtSymbol) {
+    console.log(chalk.red("YBT: symbol mismatch"));
+    process.exit(1);
+  }
```
RedFox20, line 103: This block (lines 87-103) should now be extracted into a separate function

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,245 @@
...
+  const depositMaxFee = parseDecimal(validateRange(config.pool.fees.deposit, 0, 0.05).toString(), ybtPrecision);
+  const earlyRedemptionMaxFee = parseDecimal(validateRange(config.pool.fees.earlyRedemption, 0, 0.05).toString(), ybtPrecision);
+  const matureRedemptionMaxFee = parseDecimal(validateRange(config.pool.fees.maturedRedemption, 0, 0.05).toString(), ybtPrecision);
+
+  let contractName;
+  let poolBacking;
+  let customPoolConstructorArgs = [];
+  if (config.kind === "Lido") {
+    contractName = "LidoTempusPool";
+    poolBacking = config.lido.steth;
```
RedFox20, line 118: `poolBacking` is unused

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,245 @@
...
+  } else if (config.kind === "Yearn") {
+    contractName = "YearnTempusPool";
+  } else if (config.kind === "Rari") {
+    contractName = "RariTempusPool";
+  } else {
+    console.log("No suitable protocol found");
+    process.exit(1);
+  }
```
RedFox20, line 126: This block (lines 113-126) should be refactored into a separate function such as `getContractName(config)`

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,245 @@
...
+    // Insert fundManager as first argument
+    poolConstructorArgs.splice(0, 0, config.rari.fundManager);
+  }
+
+  console.log(chalk.yellow(`${contractName} constructor arguments: `));
+  console.log(chalk.green(JSON.stringify(poolConstructorArgs)));
+  if (!(await utils.toggleConfirm("Do you confirm the constructor arguments?"))) {
+    console.log(chalk.yellow('Constructor arguments not confirmed.'));
+    process.exit(0)
+  }
```
RedFox20, line 161: This can be refactored into a function and reused in deployAmm

RedFox20: Maybe even including constructor arguments stringify

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  await utils.generateDeployment(tempusPoolContract, `${contractName}_${ybtSymbol}_maturity-${maturityTimestamp}`, network.name);
+
+  config.pool.address = tempusPoolContract.address;
+  config.pool.owner = await tempusPoolContract.owner();
+  return config;
+}
+
+async function deployAmm(config:Config, deployerPrivateKey:string) {
+  const balancerVault = config.amm.vault;
```
RedFox20, line 174: I still think pulling out some of these variables just adds extra lines, without much value. So I recommend to clean it up and use `config.amm.vault`, `config.amm.lp.name` etc etc directly

scripts/deploy.pool.config.ts
```diff
@@ -0,0 +1,245 @@
...
+    pauseWindowDuration,
+    bufferPeriodDuration,
+    owner
+  })));
+  if (!(await utils.toggleConfirm("Do you confirm the constructor arguments?"))) {
+    console.log(chalk.yellow('Constructor arguments not confirmed.'));
+    process.exit(0)
+  }
+
+  const ammConstructorArgs = [
```
RedFox20, line 201: This can be reused in the JSON.stringify() above

scripts/deploy.pool.config.ts
axic: Fixed.

scripts/deploy.pool.config.ts
axic: Fixed.

scripts/deploy.pool.config.ts
yuval-weiss: lessgo

scripts/deploy.pool.config.ts
axic: Yeah now after the refactoring it may be useful to do it. Initially I argued that we should merge the baseline (~= similar to the original scripts) version, and then do these, but at this point it is quite far from the baseline.

RedFox20 (approved):
Great work 👍🏻

PR #519, opened 2022-02-10
ci: run unit tests and integration tests in parallel

RedFox20:


run_tests.sh
```diff
@@ -0,0 +1,46 @@
+TESTS_PATH=""
```
axic, line 1:
```suggestion
#!/usr/bin/env bash
TESTS_PATH=""
```

And then there's no need to use `bash run_tests.sh`, but `./run_tests.sh`.

run_tests.sh
```diff
@@ -0,0 +1,46 @@
...
+    REPORTER="spec"
+
+  elif [[ "$POOLS" =~ "$1" ]]; then # only pool
+    ONLY_POOL="$1"
+    REPORTER="spec"
+  fi
+  shift
+done
+
+CMD="npx mocha --require hardhat/register --timeout 120000 --recursive --extension ts --reporter $REPORTER $TESTS_PATH"
```
axic, line 31: This is missing hte `--parallel` setting from `packages.json`. Is that intentional?

run_tests.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# parse args
+while [ -n "$1" ]; do
+  if [ "$1" == "integration-suite" ]; then # integration suite test
+    ENVS="$ENVS INTEGRATION=1 ETH_NODE_URI_MAINNET=$ETH_NODE_URI_MAINNET"
+    ENVS="$ENVS HARDHAT_FORK=mainnet HARDHAT_FORK_NUMBER=13533408 "
+
+  elif [ "$1" == "integration-tests" ]; then # use integration-tests folder
+    ENVS="$ENVS INTEGRATION=1 ETH_NODE_URI_MAINNET=$ETH_NODE_URI_MAINNET"
+    ENVS="$ENVS HARDHAT_FORK=mainnet HARDHAT_FORK_NUMBER=13533408 "
```
axic, line 19: I'd move the hard fork number as a config option to the top, potentially the same with fork (mainnet).

run_tests.sh
RedFox20: In fact, it needs to be `bash run_tests.sh`, so that running inside Visual Studio Code terminal it doesn't launch a new terminal instance like with `./run_tests.sh`. Not sure if it applies for Linux, but this is the case for Windows

run_tests.sh
RedFox20: --parallel is no longer needed, it never worked anyway

run_tests.sh
RedFox20: Also got this error: `'.' is not recognized as an internal or external command,
operable program or batch file.`

axic: It doesn't apply to Linux/macOS, so I'd prefer to at least have this header in the file.

run_tests.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/bash
```
axic, line 1: `#!/usr/bin/env bash` is the modern way, as it doesn't assume the actual path for bash (which usually is `/bin/bash`.

RedFox20: I'll keep it /usr/bin/bash for compatibility sake with the fake environment that Windows uses to run this stack. Compatibility is more important

axic: > Compatibility is more important

I do not have bash at `/usr/bin/bash`:
```sh
$ ls /bin/bash
/bin/bash
$ ls /usr/bin/bash
ls: /usr/bin/bash: No such file or directory
```

RedFox20: Changed it to `#!/bin/bash`, but unless `cross-env` is in global PATH, it can't be run outside of npm scripts, at least on windows

run_tests.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+
+TESTS_PATH=""
+ETH_NODE_URI_MAINNET="https://eth-mainnet.alchemyapi.io/v2/yInmJCdHvCzXe4qAxfPMDFgrbYIGsUhJ"
```
mijovic, line 4: we can't leave this in the public repo...
this needs to be added every time one runs the script instead

mijovic: AlchemyApi has free tier limit, and if someone wants to play with us could easily drain it quickly every time we change it

RedFox20: Alright. I added the fix 👍🏻

mijovic (approved):

PR #521, opened 2022-02-15
Make AMP parameter decimal 

StefanVitor:


contracts/amm/TempusAMM.sol
```diff
@@ -693,26 +697,24 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
     }
 
     function _startAmplificationParameterUpdate(uint256 rawEndValue, uint256 endTime) private {
```
yuval-weiss, line 699: If we want to stick to the current naming convention regarding raw vs non-raw amplification rate, `uint256 rawEndValue` should be changed to `uint256 endValue`. With the current naming convention, `raw` amp is the pre-scaled value.

contracts/amm/TempusAMM.sol
```diff
@@ -120,8 +124,9 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
     {
         assert(_TOTAL_TOKENS == 2);
 
-        _require(amplificationStart >= _MIN_AMP, Errors.MIN_AMP);
-        _require(amplificationStart <= _MAX_AMP, Errors.MAX_AMP);
+        uint256 initialAmp = Math.mul(amplificationStart, _AMP_PRECISION);
```
yuval-weiss, line 127: We should also allow providing a decimal (non-raw) amplification factor in the constructor. With the current implementation the `TempusAMM::startAmplificationParameterUpdate` expected non-raw amp value (min value 1000) while the constructor still accepts raw amp values (min value 1).

contracts/amm/TempusAMM.sol
mijovic: Probably we need to rename this, I agree

contracts/amm/TempusAMM.sol
```diff
@@ -688,22 +691,20 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
      * NOTE: Internally, the amplification parameter is represented using higher precision. The values returned by
```
yuval-weiss, line 691: we should change this note or remove it completely. Also, the `@dev` section above it needs to be updated.

test/utils/TempusAMM.ts
```diff
@@ -229,32 +231,38 @@ export class TempusAMM extends ContractBase {
...
+    const endTime = this.startedAmpUpdateTime + Math.trunc(ampDiff / +ampParam.precision) * oneAmpUpdateTime;
+
+    return this.contract.startAmplificationParameterUpdate(this.targetAmp , endTime);
   }
 
   async forwardToAmplification(ampValue: number): Promise<void> {
     let targetTimestamp: number;
-    if (this.startAmp == ampValue) {
+    const ampParam = await this.getAmplificationParam();
+    const ampValueWithPrecision = Math.trunc(+ampValue * +ampParam.precision);
```
yuval-weiss, line 251: We can rename the `ampValue` argument to `rawAmpValue` if we want to stick to the Balancer terminology -- that way we can get rid of the "WithPrecision" suffix of `ampValueWithPrecision` which imo is nicer.

Also this can probably be applied to other methods in this file (such as `startAmplificationUpdate`) such that the function arguments are prefixed with `raw` which makes it easier to understand what's scaled and what isn't.

test/utils/TempusAMM.ts
```diff
@@ -69,8 +71,8 @@ export class TempusAMM extends ContractBase {
       "Tempus LP token", 
       "LP", 
       tempusPool.address,
-      amplificationStart,
-      amplificationEnd,
+      +amplificationStart * AMP_PRECISION,
```
yuval-weiss, line 74: we can add `raw` here too :)

mijovic (commented):
This looks good to go. Can you just squash commits into one probably?

mijovic (approved):

PR #522, opened 2022-02-15
refactor etherscan deployment script 

yuval-weiss:
* The script now works with all protocols
* Minimized prompt-provided data - instead of providing the deployment constructor arguments, the script now prompts for the contract deployment tx ids and extracts the constructor args from the txs' input.


mijovic (approved):

yuval-weiss:
Resolved #520

PR #527, opened 2022-02-17
Replace pool with shares in AMM

axic:


contracts/amm/TempusAMM.sol
```diff
@@ -210,16 +211,15 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
         external
         view
         returns (uint256 lpTokens)
     {
         (IERC20[] memory ammTokens, uint256[] memory balances, ) = getVault().getPoolTokens(getPoolId());
         uint256[] memory amountsOut = new uint256[](2);
-        (amountsOut[0], amountsOut[1]) = (address(ammTokens[0]) == address(tempusPool.principalShare()))
-            ? (principalsStaked, yieldsStaked)
-            : (yieldsStaked, principalsStaked);
+        amountsOut[0] = token0Staked;
```
RedFox20, line 221: Does the caller always easily know which one is token0 and which is token1? Or should we follow previous pattern:
```ts
(amountsOut[0], amountsOut[1]) = (address(ammTokens[0]) == address(token0Staked))
            ? (token0Staked, token1Staked)
            : (token1Staked, token0Staked);
```

RedFox20, line 221: How does the caller know which one is token0 and which is token1 ?

axic: Need to make `token0`/`token1` public in the AMM so the caller can query. It may not be yields/pcincipals only, it could be principals/principals -- at least that's what I recall being said.

PR #528, opened 2022-02-17
[TempusController] Remove some redundant functions

yuval-weiss:


contracts/TempusController.sol
```diff
@@ -232,39 +271,6 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
         vault.swap(singleSwap, fundManagement, minReturn, deadline);
     }
 
-    function _depositAndProvideLiquidity(
```
yuval-weiss, line 235: It's only used by the external `depositAndProvideLiquidity` (which contains no logic other than the `_depositAndProvideLiquidity` call) so imo better to not have this as a separate function.

contracts/TempusController.sol
```diff
@@ -508,38 +502,6 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
         vault.exitPool(poolId, sender, payable(recipient), request);
     }
 
-    function _exitAmmGivenAmountsOutAndEarlyRedeem(
```
yuval-weiss, line 511: same as what I described in the  `_depositAndProvideLiquidity` comment

contracts/TempusController.sol
```diff
@@ -177,16 +225,7 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
...
-            require(tempusAMM.transferFrom(msg.sender, address(this), lpTokens), "LP token transfer failed");
-
-            // exit amm and sent shares to controller
-            uint256[] memory minAmountsOutFromLP = getAMMOrderedAmounts(
-                tempusAMM,
-                minPrincipalsStaked,
-                minYieldsStaked
-            );
-            _exitTempusAMMGivenLP(tempusAMM, address(this), address(this), lpTokens, minAmountsOutFromLP, false);
+            _exitTempusAMM(tempusAMM, lpTokens, minPrincipalsStaked, minYieldsStaked, address(this), false);
```
yuval-weiss, line 228: `_exitTempusAMM` contained some logic that was duplicated here

contracts/TempusController.sol
```diff
@@ -26,7 +26,7 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
     /// Registry for valid pools and AMM's to avoid fake address injection
     mapping(address => bool) private registry;
 
-    constructor() Versioned(1, 1, 0) {}
+    constructor() Versioned(1, 1, 1) {}
```
axic, line 29: I don't think there's a need to change any version as this is only internal refactoring. External interface nor semantics are changing.

contracts/TempusController.sol
RedFox20: Hmm, I think this change alone should be in its own PR so it can be reviewed properly. I understand you added a `recipient` param to _exitTempusAMM and thus changed some of this logic, so now it needs to be properly reviewed to ensure it is exactly the same.

contracts/TempusController.sol
mijovic: Totally agree with this

RedFox20:
@yuval-weiss This PR looks all good, lets get it merged asap 😃

RedFox20 (approved):

PR #523, opened 2022-02-15
amm: Rename amplificationStart/amplificationEnd to amplificationStartValue/amplificationEndValue

axic:
It is easy to confuse it with timestamps otherwise.

PR #531, opened 2022-02-18
Remove dependancy between TempusAMM and TempusPool

StefanVitor:


contracts/ITempusController.sol
```diff
@@ -62,6 +62,7 @@ interface ITempusController {
     /// @param isBackingToken specifies whether the deposited asset is the Backing Token or Yield Bearing Token
     function depositAndProvideLiquidity(
         ITempusAMM tempusAMM,
+        ITempusPool tempusPool,
```
mijovic, line 65: this will need NatSpec docs as well

contracts/TempusController.sol
```diff
@@ -304,38 +318,38 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
...
 
-        (uint256 mintedShares, uint256 depositedBT, uint256 fee, uint256 rate) = targetPool.onDepositYieldBearing(
+        (uint256 mintedShares, uint256 depositedBT, uint256 fee, uint256 rate) = tempusPool.onDepositYieldBearing(
             transferredYBT,
             recipient
         );
 
         emit Deposited(
-            address(targetPool),
+            address(tempusPool),
```
mijovic, line 352: Ah, this refactor is so annoying, it might be better if it was done in separate PR. Probably next time

contracts/amm/TempusAMM.sol
```diff
@@ -98,9 +95,10 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
         IVault vault,
         string memory name,
         string memory symbol,
-        ITempusPool pool,
-        uint256 amplificationStart,
-        uint256 amplificationEnd,
+        IPoolShare[2] memory tokens,
```
mijovic, line 98: Is it an array because of the `stack too deep` error? I would try having two separate variables instead of array if possible

contracts/amm/TempusAMM.sol
```diff
@@ -179,31 +170,35 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-                uint256 newYields = yieldsIn ? (yields - amountIn) : (yields + amountOut);
-                uint256 newDifference = (newPrincipals > newYields)
-                    ? (newPrincipals - newYields)
-                    : (newYields - newPrincipals);
+                uint256 amountOut = getExpectedReturnGivenIn(amountIn, tokenIn);
+                uint256 newToken0Amount = (tokenIn == token1) ? (token0Amount + amountOut) : (token0Amount - amountIn);
+                uint256 newToken1Amount = (tokenIn == token1) ? (token1Amount - amountIn) : (token1Amount + amountOut);
+                uint256 newDifference = (newToken0Amount > newToken1Amount)
+                    ? (newToken0Amount - newToken1Amount)
+                    : (newToken1Amount - newToken0Amount);
```
mijovic, line 199: Ićll need to double-check this in next round of review

contracts/amm/TempusAMM.sol
```diff
@@ -213,16 +208,16 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
     }
 
     // NOTE: Return value in AMM decimals precision (1e18)
-    function getExpectedBPTInGivenTokensOut(uint256 principalsStaked, uint256 yieldsStaked)
+    function getExpectedBPTInGivenTokensOut(uint256 token0Staked, uint256 token1Staked)
```
mijovic, line 211: we need to double-check if this method is used at all. If not, I would revert as unsupported action. However, even if we leave it, we need to rename params to `token0Out` and `token1Out`

contracts/amm/TempusAMM.sol
```diff
@@ -250,14 +245,14 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
     function getExpectedTokensOutGivenBPTIn(uint256 bptAmountIn)
         external
         view
-        returns (uint256 principals, uint256 yields)
+        returns (uint256 token0Amount, uint256 token1Amount)
```
mijovic, line 248:
```suggestion
        returns (uint256 token0Out, uint256 token1Out)
```

contracts/amm/TempusAMM.sol
```diff
@@ -250,14 +245,14 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
     {
         // We don't need to scale balances down here
         // as calculation for amounts out is based on btpAmountIn / totalSupply() ratio
         // Adjusting balances with rate, and then undoing it would just cause additional calculations
-        (, uint256[] memory balances, ) = getVault().getPoolTokens(getPoolId());
+        (IERC20[] memory ammTokens, uint256[] memory balances, ) = getVault().getPoolTokens(getPoolId());
         uint256[] memory amountsOut = StableMath._calcTokensOutGivenExactBptIn(balances, bptAmountIn, totalSupply());
-        (principals, yields) = (address(_token0) == address(tempusPool.principalShare()))
+        (token0Amount, token1Amount) = (address(token0) == address(ammTokens[0]))
```
mijovic, line 255: I think this check is redundant as ammTokens are also sorted, so `ammTokens[0]` is always `token0`

contracts/amm/TempusAMM.sol
```diff
@@ -430,7 +425,7 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
         uint256[] memory scalingFactors,
         uint256[] memory tokenRates,
         bytes memory userData
-    ) private returns (uint256 bptAmountOut, uint256[] memory amountsIn) {
+    ) private view returns (uint256 bptAmountOut, uint256[] memory amountsIn) {
```
mijovic, line 428: How come this compiles? This is not a view function, at least I don't see it like that. It updates the state

contracts/amm/TempusAMM.sol
```diff
@@ -523,7 +518,7 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
         uint256[] memory scalingFactors,
         uint256[] memory tokenRates,
         bytes memory userData
-    ) private returns (uint256, uint256[] memory) {
+    ) private view returns (uint256, uint256[] memory) {
```
mijovic, line 521: Also here

contracts/stats/Stats.sol
```diff
@@ -65,32 +65,32 @@ contract Stats is ChainlinkTokenPairPriceFeed, Versioned {
...
     /// Gets the estimated amount of Principals and Yields after a successful deposit
-    /// @param pool Which tempus pool
+    /// @param tempusPool Which tempus pool
     /// @param amount Amount of BackingTokens or YieldBearingTokens that would be deposited
     /// @param isBackingToken If true, @param amount is in BackingTokens, otherwise YieldBearingTokens
     /// @return Amount of Principals (TPS) and Yields (TYS) in Principal/YieldShare decimal precision.
     ///         TPS and TYS are minted in 1:1 ratio, hence a single return value.
     function estimatedMintedShares(
-        ITempusPool pool,
+        ITempusPool tempusPool,
```
mijovic, line 74: this Stats refactor is a bit unrelated, can you remove it and do later in separate PR?

mijovic:
Needs rebase because #523 was merged

contracts/amm/TempusAMM.sol
StefanVitor: Yes, exactly ("CompilerError: Stack too deep when compiling inline assembly: Variable headStart is 1 slot(s) too deep inside the stack. Error HH600: Compilation failed"), so I decided on this way to prevent that error.

contracts/amm/TempusAMM.sol
StefanVitor: As I see, it's not update any state.

contracts/amm/TempusAMM.sol
StefanVitor: As I see, it's not update any state (and compile pass)

contracts/amm/TempusAMM.sol
mijovic: I am thinking if it is worth packing amplification data to struct and use it like that. We could do this as separate PR, and then rebase this one. In that case we could use two addresses instead of array

StefanVitor: Agree with you, I'll put data for amplification into a struct (first, I'll change that, create PR for that, and then rebase and continue with this PR).

contracts/ITempusController.sol
RedFox20, line 65: This file looks all good ✅

package.json
```diff
@@ -49,6 +49,7 @@
   },
   "scripts": {
     "build": "npx hardhat compile",
+    "clean": "rm -rf ./cache && rm -rf ./artifacts",
```
RedFox20, line 52: Nice, I'm surprised we didn't add this before 😆

contracts/amm/interfaces/ITempusAMM.sol
```diff
@@ -30,37 +32,36 @@ interface ITempusAMM {
 
     /// Calculates the expected returned swap amount
     /// @param amount The given input amount of tokens
-    /// @param yieldShareIn Specifies whether to calculate the swap from TYS to TPS (if true) or from TPS to TYS
+    /// @param tokenIn Specifies inToken pool share address
```
RedFox20, line 35: `Specifies which token [token0 or token1] that @param amount refers to`

Also, here we have security issue, if malicious operator passes a random IPoolShare address which is deployed by them, what happens @StefanVitor ?

contracts/amm/TempusAMM.sol
```diff
@@ -160,13 +153,11 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-    function getExpectedReturnGivenIn(uint256 amount, bool yieldShareIn) public view returns (uint256) {
+    function getExpectedReturnGivenIn(uint256 amount, IPoolShare tokenIn) public view returns (uint256) {
         (, uint256[] memory balances, ) = getVault().getPoolTokens(getPoolId());
         (uint256 currentAmp, ) = _getAmplificationParameter();
-        (IPoolShare tokenIn, IPoolShare tokenOut) = yieldShareIn
-            ? (tempusPool.yieldShare(), tempusPool.principalShare())
-            : (tempusPool.principalShare(), tempusPool.yieldShare());
-        (uint256 indexIn, uint256 indexOut) = address(tokenIn) == address(_token0) ? (0, 1) : (1, 0);
+        IPoolShare tokenOut = (tokenIn == token0) ? token1 : token0;
+        (uint256 indexIn, uint256 indexOut) = address(tokenIn) == address(token0) ? (0, 1) : (1, 0);
```
RedFox20, line 160: Aha, ok, so it's validated here. Also, did Djordje's suggestions of `tokenIn == token0` not work?

contracts/amm/interfaces/ITempusAMM.sol
StefanVitor: Good point, I'll check that tokenIn must be token0 or token1 (and with that, it should be all fine).

test/utils/TempusController.ts
```diff
@@ -251,13 +253,14 @@ export class TempusController extends ContractBase {
 
     return this.connect(user).exitAmmGivenLpAndRedeem(
       amm.address,
+      pool.tempus.address,
       amm.toBigNum(lpTokens),
-      amm.principalShare.toBigNum(principals),
-      amm.yieldShare.toBigNum(yields),
+      amm.token0.toBigNum(principals),
```
RedFox20, line 258: Need to change `principals` and `yields` here too

contracts/TempusController.sol
```diff
@@ -560,16 +589,16 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
...
+        if (!targetPool.matured()) {
             if (((yields > principals) ? (yields - principals) : (principals - yields)) >= maxLeftoverShares) {
-                (uint256 swapAmount, bool yieldsIn) = tempusAMM.getSwapAmountToEndWithEqualShares(
-                    principals,
-                    yields,
-                    maxLeftoverShares
-                );
+                (uint256 swapAmount, IPoolShare tokenIn) = (targetPool.principalShare() == tempusAMM.token0())
+                    ? tempusAMM.getSwapAmountToEndWithEqualShares(principals, yields, maxLeftoverShares)
+                    : tempusAMM.getSwapAmountToEndWithEqualShares(yields, principals, maxLeftoverShares);
```
RedFox20, line 596: I think this can now be refactored as follows:
```suggestion
                (uint256 amount0, uint256 amount1) = (targetPool.principalShare() == tempusAMM.token0())
                    ? (principals, yields)
                    : (yields, principals);
                (uint256 swapAmount, IPoolShare tokenIn) = tempusAMM.getSwapAmountToEndWithEqualShares(
                    amount0, amount1, maxLeftoverShares
                );
```
It is slightly cleaner and we use this `amount0, amount1` pattern already in other functions 👌🏻 
Also see `getAMMOrderedAmounts()` function

StefanVitor: In this way, we will use two more variables and get a "Stack too deep, try removing local variables." error.

RedFox20: That's unfortunate. If no vars can be removed, then can't change this

contracts/amm/TempusAMM.sol
```diff
@@ -127,31 +125,26 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
 
         // Immutable variables cannot be initialized inside an if statement, so we must do conditional assignments
-        (IPoolShare token0, IPoolShare token1) = yieldShare < principalShare
-            ? (yieldShare, principalShare)
-            : (principalShare, yieldShare);
-        (_token0, _token1) = (token0, token1);
+        (IPoolShare token0Local, IPoolShare token1Local) = tokens[0] > tokens[1]
+            ? (tokens[1], tokens[0])
+            : (tokens[0], tokens[1]);
+        (token0, token1) = (token0Local, token1Local);
```
RedFox20, line 138: Can you try without local variables? I see they are not used outside of this expression.
Something like this:
```suggestion
        (token0, token1) = tokens[0] > tokens[1] ? (tokens[1], tokens[0]) : (tokens[0], tokens[1]);
```

RedFox20: Never mind, I see that you use them below

contracts/amm/TempusAMM.sol
```diff
@@ -179,31 +172,33 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-                uint256 newPrincipals = yieldsIn ? (principals + amountOut) : (principals - amountIn);
-                uint256 newYields = yieldsIn ? (yields - amountIn) : (yields + amountOut);
-                uint256 newDifference = (newPrincipals > newYields)
-                    ? (newPrincipals - newYields)
-                    : (newYields - newPrincipals);
+                uint256 amountOut = getExpectedReturnGivenIn(amountIn, tokenIn);
+                uint256 newToken0Amount = (tokenIn == token1) ? (token0Amount + amountOut) : (token0Amount - amountIn);
+                uint256 newToken1Amount = (tokenIn == token1) ? (token1Amount - amountIn) : (token1Amount + amountOut);
+                uint256 newDifference = (newToken0Amount > newToken1Amount)
+                    ? (newToken0Amount - newToken1Amount)
```
RedFox20, line 198: Maybe we should start thinking about Math utils like `Math.Abs` and `Math.Difference`. Not sure how good the inliner will be though to eliminate the pure calls

contracts/amm/TempusAMM.sol
```diff
@@ -838,20 +829,22 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
+    {
         IERC20[] memory tokens = new IERC20[](_TOTAL_TOKENS);
-        IPoolShare yieldShare = pool.yieldShare();
-        IPoolShare principalShare = pool.principalShare();
-        (tokens[0], tokens[1]) = (yieldShare < principalShare)
-            ? (IERC20(address(yieldShare)), IERC20(address(principalShare)))
-            : (IERC20(address(principalShare)), IERC20(address(yieldShare)));
+        (tokens[0], tokens[1]) = (token0Param > token1Param)
+            ? (IERC20(address(token1Param)), IERC20(address(token0Param)))
+            : (IERC20(address(token0Param)), IERC20(address(token1Param)));
```
RedFox20, line 847: We do this sorting so often, wish we had some common utilities for it. Maybe in some other PR...

contracts/amm/interfaces/ITempusAMM.sol
```diff
@@ -30,37 +32,36 @@ interface ITempusAMM {
...
-    /// @dev Returns amount that user needs to swap to end up with almost the same amounts of Principals and Yields
-    /// @param principals User's Principals balance
-    /// @param yields User's Yields balance
-    /// @param threshold Maximum difference between final balances of Principals and Yields
-    /// @return amountIn Amount of Principals or Yields that user needs to swap to end with almost equal amounts
-    /// @return yieldsIn Specifies the "direction" of the swap -
-    ///         whether Yields should be swapped for Principals (yieldsIn=true) or vice versa
+    /// @dev Returns amount that user needs to swap to end up with almost the same amounts of Token0 and Token1
+    /// @param token0Amount User's Token0 balance
+    /// @param token1Amount User's Token1 balance
```
RedFox20, line 41:
```suggestion
    /// @param token0Amount Desired token0 amount after swap()
    /// @param token1Amount Desired token1 amount after swap()
```

contracts/stats/Stats.sol
```diff
@@ -173,30 +179,35 @@ contract Stats is ChainlinkTokenPairPriceFeed, Versioned {
         )
     {
         if (lpTokens > 0) {
-            (principalsStaked, yieldsStaked) = tempusAMM.getExpectedTokensOutGivenBPTIn(lpTokens);
+            {
+                (uint256 principalsOut, uint256 yieldsOut) = tempusAMM.getExpectedTokensOutGivenBPTIn(lpTokens);
```
RedFox20, line 183: Names of `principalsOut` and `yieldsOut` no longer applies, should be `token0` and `token1`, right?

Could we maybe add a Library for dealing with TempusAMM, this manual comparison for token0() is barbaric and leads to maintenance hell.

```ts
    using TempusPoolAMMUtils for ITempusAMM;
```

```ts
library TempusPoolAMMUtils {
    function getOrderedPYAmounts(ITempusAMM amm, ITempusPool pool, uint256 amount0, uint256 amount1) internal view returns (uint256 principalsAmount, uint256 yieldsAmount) {
        (principalsAmount, yieldsAmount) = (pool.principalShare() == amm.token0())
                    ? (amount0, amount1)
                    : (amount1, amount0);
    }
}
```

contracts/stats/Stats.sol
```diff
@@ -173,30 +179,35 @@ contract Stats is ChainlinkTokenPairPriceFeed, Versioned {
...
         // before maturity we need to have equal amount of shares to redeem
-        if (!tempusAMM.tempusPool().matured()) {
-            (uint256 amountIn, bool yieldsIn) = tempusAMM.getSwapAmountToEndWithEqualShares(
-                principals,
-                yields,
-                threshold
-            );
-            uint256 amountOut = (amountIn != 0) ? tempusAMM.getExpectedReturnGivenIn(amountIn, yieldsIn) : 0;
+        if (!pool.matured()) {
+            (uint256 amountIn, IPoolShare tokenIn) = (pool.principalShare() == tempusAMM.token0())
```
RedFox20, line 194: And for the reverse case here, could try introducing a second helper:
```ts
library TempusPoolAMMUtils {
    function getOrderedAmmAmounts(
        ITempusAMM amm,
        ITempusPool pool,
        uint256 principals,
        uint256 yields
    ) internal view returns (
        uint256 token0Amount,
        uint256 token1Amount
    ) {
        (token0Amount, token1Amount) = (pool.principalShare() == amm.token0())
            ? (principals, yields)
            : (yields, principals);
    }
}
```

contracts/stats/Stats.sol
```diff
@@ -226,20 +238,23 @@ contract Stats is ChainlinkTokenPairPriceFeed, Versioned {
...
         bool toBackingToken
     ) public view returns (uint256 tokenAmount, uint256 lpTokensRedeemed) {
-        require(!tempusAMM.tempusPool().matured(), "Pool already finalized!");
+        require(!pool.matured(), "Pool already finalized!");
 
         if (principalsStaked > 0 || yieldsStaked > 0) {
-            lpTokensRedeemed = tempusAMM.getExpectedBPTInGivenTokensOut(principalsStaked, yieldsStaked);
+            lpTokensRedeemed = (pool.principalShare() == tempusAMM.token0())
+                ? tempusAMM.getExpectedBPTInGivenTokensOut(principalsStaked, yieldsStaked)
+                : tempusAMM.getExpectedBPTInGivenTokensOut(yieldsStaked, principalsStaked);
```
RedFox20, line 253: If we had the new utilities here, it could be:

```suggestion
            (uint256 token0Amt, uint256 token1Amt) = getOrderedAmmAmounts(this, pool, principalsStaked, yieldsStaked);
            lpTokensRedeemed = tempusAMM.getExpectedBPTInGivenTokensOut(token0Amt, token1Amt);
```

test/utils/TempusAMM.ts
```diff
@@ -102,19 +103,19 @@ export class TempusAMM extends ContractBase {
     return this.fromBigNum(await this.contract.getRate());
   }
 
-  async getExpectedReturnGivenIn(inAmount: NumberOrString, yieldShareIn: boolean) : Promise<NumberOrString> {
-    return this.principalShare.fromBigNum(await this.contract.getExpectedReturnGivenIn(this.principalShare.toBigNum(inAmount), yieldShareIn));
+  async getExpectedReturnGivenIn(inAmount: NumberOrString, tokenIn: PoolShare) : Promise<NumberOrString> {
+    return this.token0.fromBigNum(await this.contract.getExpectedReturnGivenIn(this.token0.toBigNum(inAmount), tokenIn.address));
```
RedFox20, line 107: For correctness sake, I think this should be `tokenIn.fromBigNum`, and `tokenIn.toBigNum(inAmount)` now.

test/utils/TempusAMM.ts
```diff
@@ -185,8 +186,8 @@ export class TempusAMM extends ContractBase {
     const poolId = await this.contract.getPoolId();
     
     const assets = [
-      { address: this.principalShare.address, amountOut: this.principalShare.toBigNum(amountsOut[0]) },
-      { address: this.yieldShare.address, amountOut: this.principalShare.toBigNum(amountsOut[1]) }
+      { address: this.token0.address, amountOut: this.token0.toBigNum(amountsOut[0]) },
+      { address: this.token1.address, amountOut: this.token0.toBigNum(amountsOut[1]) }
```
RedFox20, line 190: Looks like there was a mistake in old code here. It should be `this.token1.toBigNum` instead of `token0`

test/utils/TempusPoolAMM.ts
```diff
@@ -0,0 +1,100 @@
...
+    controller: TempusController,
+    principalShare: PoolShare,
+    yieldShare: PoolShare,
+    rawAmplificationStart: number,
+    rawAmplificationEnd: number,
+    amplificationEndTime: number,
+    swapFeePercentage: number
+  ): Promise<TempusPoolAMM> {
+    const [sender] = new MockProvider().getWallets();
+    const mockedWETH = await deployMockContract(sender, WETH_ARTIFACTS.abi);
```
RedFox20, line 37: Wait, why are we creating a mocked WETH contract here? I don't remember what kind of requirements there was for this.

But at least in the `create()` function, we should somehow put a comment that this is for testing only, or change the name to `createMock()` ?

Again, I don't remember anymore what the original architecture for this thing was...

test/utils/TempusPoolAMM.ts
```diff
@@ -0,0 +1,100 @@
...
+    await controller.register(owner, tempusAMM.address);
+    return new TempusPoolAMM(tempusAMM, vault, principalShare, yieldShare);
+  }
+  
+  async getExpectedTokensOutGivenBPTIn(inAmount: NumberOrString): Promise<{token0Out:number, token1Out:number}> {
+    const p = await super.getExpectedTokensOutGivenBPTIn(inAmount);
+    const [principalsOut, yieldsOut] = (this.principalShare.address == this.token0.address) 
+        ? [p.token0Out, p.token1Out]
+        : [p.token1Out, p.token0Out];
+    return {token0Out: +principalsOut, token1Out: +yieldsOut};
```
RedFox20, line 67: The output object variables should be `principalsOut: +principalsOut, yieldsOut: +yieldsOut`, because this is converting from Token0/1 into PrincipalsYields

RedFox20: To make it easier, use utility functions to avoid repeating code:
```ts
public getPYAmount(token0Amount: NumberOrString, token1Amount: NumberOrString): {principalsOut:number, yieldsOut:number} {
  return (this.principalShare.address == this.token0.address)
    ? {principalsOut: +token0Amount, yieldsOut: +token1Amount}
    : {principalsOut: +token1Amount, yieldsOut: +token0Amount};
}
```

test/utils/TempusPoolAMM.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  async getExpectedLPTokensForTokensIn(principalsAmountIn:NumberOrString, yieldsAmountIn:NumberOrString): Promise<NumberOrString> {
+    const [token0AmountIn, token1AmountIn] = (this.principalShare.address == this.token0.address) 
+        ? [principalsAmountIn, yieldsAmountIn]
+        : [yieldsAmountIn, principalsAmountIn];
```
RedFox20, line 73: And for this one, you could use an inverse utility:
```ts
public getAmmAmounts(principalsAmount: NumberOrString, yieldsAmount: NumberOrString): {token0Out:number, token1Out:number} {
    return (this.principalShare.address == this.token0.address)
      ? {token0Out: +principalsAmount, token1Out: +yieldsAmount}
      : {token0Out: +yieldsAmount, token1Out: +principalsAmount};
}
```

RedFox20:
```ts
    const a = this.getAmmAmounts(principalsAmountIn, yieldsAmountIn);
    return super.getExpectedLPTokensForTokensIn(a.token0Out, a.token1Out);

File: contracts/stats/Stats.sol (hunk 20 above)
Comment (StefanVitor) [reply contracts/stats/Stats.sol:183]: As I understand Djordje (and it also looks logical to me), only in TempusAMM we will not use principals and yields because we remove dependency from TempusPool (we will use token0 and token1 only there). For other contracts such as Stats and TempusController, they are still working with TempusPool, so on those contracts, it should continue to stay principals and yields (because there is connection still with TempusPool from that contracts).

Comment (StefanVitor) [reply contracts/stats/Stats.sol:183]: And for TempusPoolAMMUtils, I'll add that.

File: test/amm/TempusAMM.test.ts [hunk 28]
```diff
@@ -350,13 +351,13 @@ describeForEachPool("TempusAMM", (testFixture:PoolTestFixture) =>
     await createPools({yieldEst:0.1, duration:ONE_YEAR*300, amplifyStart:1, amplifyEnd:95, ammBalancePrincipal: 10000, ammBalanceYield: 100000});
 
     // basic swap with Interest Rate aligned to balances with increasing amplification
-    await checkSwap(owner, {amplification: 5, swapAmountIn: 1, swapAmountOut: 9.800039358937214, principalIn: true});
```
RedFox20, line 353: Maybe should have left `principalIn: boolean` unchanged and handle it inside checkSwap? Caused a lot of extra work...

test/amm/TempusAMM.test.ts
```diff
@@ -135,8 +136,8 @@ describeForEachPool("TempusAMM", (testFixture:PoolTestFixture) =>
     await testFixture.amm.exitPoolExactLpAmountIn(owner, inputAmount);
     const balancePrincipalsAfter = +await testFixture.principals.balanceOf(owner);
     const balanceYieldsAfter = +await testFixture.yields.balanceOf(owner);
-    expect(balancePrincipalsBefore + expectedReturn.principals).to.be.within(0.999999 * balancePrincipalsAfter, 1.0000001 * balancePrincipalsAfter);
-    expect(balanceYieldsBefore + expectedReturn.yields).to.be.within(0.999999 * balanceYieldsAfter, 1.0000001 * balanceYieldsAfter);
+    expect(balancePrincipalsBefore + expectedReturn.token0Out).to.be.within(0.999999 * balancePrincipalsAfter, 1.0000001 * balancePrincipalsAfter);
+    expect(balanceYieldsBefore + expectedReturn.token1Out).to.be.within(0.999999 * balanceYieldsAfter, 1.0000001 * balanceYieldsAfter);
```
RedFox20, line 140: I guess this `token0Out` and `token1Out` is the same mistake I pointed out in TempusPoolAMM.ts, it should be `expectedReturn.principalsOut` or something like that

contracts/stats/Stats.sol
StefanVitor: When I use library and function getOrderedPYAmounts, I got 

CompilerError: Stack too deep, try removing local variables.
   --> contracts/stats/Stats.sol:211:52:
    |
211 |                 (principalsStaked, yieldsStaked) = tempusAMM.getOrderedPYAmounts(pool, principalsOut, yieldsOut);
    |                                                    ^^^^^^^^^


Error HH600: Compilation failed

test/utils/TempusPoolAMM.ts
StefanVitor: Created function in TempusAMM test, so in TempusPoolAMM, I'll only use this function.

 static async createMock(): Promise<MockContract> {
    const [sender] = new MockProvider().getWallets();
    const mockedWETH = await deployMockContract(sender, WETH_ARTIFACTS.abi);
    return mockedWETH;
  }

And for purpose, this mocked WETH is used only for deploying vault.

contracts/amm/TempusAMM.sol
StefanVitor: I saw that something like `Math.Abs` could be used on a few places in different contracts, so in my opinion, it's best that this should be add in separate PR.

contracts/amm/TempusAMM.sol
StefanVitor: Yes, I agree, but where is this sorting other use, I saw that only in `TempusAMM.sol`?

.solcover.js
```diff
@@ -7,5 +7,6 @@ module.exports = {
   mocha: {
     grep: "@skip-on-coverage", // Find everything with this tag
     invert: true               // Run the grep's inverse set.
-  }
+  },
+  configureYulOptimizer: true
```
mijovic, line 11: Is this accidentally here?

RedFox20 (approved):

.solcover.js
StefanVitor: On code-coverage, we got "stack too deep" error in TempusAMM constructor. I read "code-coverage" documentation, and they have a property with a name configureYulOptimizer (explanation for this property is "(Experimental) Setting to true should resolve "stack too deep" compiler errors in large projects using ABIEncoderV2"). After I set this property in .solcover.js, everything passed fine. Maybe you have some better solution?

RedFox20 (approved):

PR #532, opened 2022-02-21
amm: Add back swapGivenOut method.

mijovic:
TODO:
- [x] Add tests

PR #533, opened 2022-02-21
controller: Adding leveraged Yield exposure via single transaction

mijovic:
Depends on #532 
TODO:
- [x] add tests

### DIFF
--- a/contracts/ITempusController.sol
+++ b/contracts/ITempusController.sol
@@ -90,6 +90,25 @@ interface ITempusController {
         uint256 deadline
     ) external payable returns (uint256);
 
+    /// @dev Atomically deposits YBT/BT to TempusPool and swaps Capitals for Yields to get leveraged exposure to yield
+    /// @param tempusPool TempusPool to be used for depositing YBT/BT
+    /// @param tempusAMM TempusAMM to use to swap Capitals for Yields
+    /// @param leverageMultiplier Multiplier to use for leverage, 18 decimal precision. In case of 2x leverage pass 2e18
+    /// @param tokenAmount Amount of YBT/BT to be deposited in underlying YBT/BT decimal precision
+    /// @param isBackingToken specifies whether the deposited asset is the Backing Token or Yield Bearing Token
+    /// @param maxCapitalsRate Maximum exchange rate of Capitals (denominated in Yields) when getting Yield in return
+    /// @param deadline A timestamp by which the transaction must be completed, otherwise it would revert
+    /// @return Amount of Capitals and Yields transferred to `msg.sender`
+    function depositAndLeverage(
+        ITempusPool tempusPool,
+        ITempusAMM tempusAMM,
+        uint256 leverageMultiplier,
+        uint256 tokenAmount,
+        bool isBackingToken,
+        uint256 maxCapitalsRate,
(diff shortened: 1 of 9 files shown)


contracts/ITempusController.sol
```diff
@@ -90,6 +90,25 @@ interface ITempusController {
...
+    /// @dev Atomically deposits YBT/BT to TempusPool and swaps Capitals for Yields to get leveraged exposure to yield
+    /// @param tempusPool TempusPool to be used for depositing YBT/BT
+    /// @param tempusAMM TempusAMM to use to swap Capitals for Yields
+    /// @param leverageMultiplier Multiplier to use for leverage, 18 decimal precision. In case of 2x leverage pass 2e18
+    /// @param tokenAmount Amount of YBT/BT to be deposited in underlying YBT/BT decimal precision
+    /// @param isBackingToken specifies whether the deposited asset is the Backing Token or Yield Bearing Token
+    /// @param maxCapitalsRate Maximum exchange rate of Capitals (denominated in Yields) when getting Yield in return
+    /// @param deadline A timestamp by which the transaction must be completed, otherwise it would revert
+    /// @return Amount of Capitals and Yields transferred to `msg.sender`
+    function depositAndLeverage(
```
yuval-weiss, line 102: I think we can just merge `depositAndFix` and `depositAndLeverage` into one function (maybe `depositAndSwap`).... It's basically the same but different swap direction

yuval-weiss: for this, we can replace `leverageMultiplier` with an argument that specifies the percentage of the minted shares that should be swapped (swap direction would be another variable) -- e.g. `yieldsIn=false` and `swapPercentage=1e18` for fixing your yield.

mijovic: I agree this is the easiest solution, but it doesn't allow exiting with exactly 2x more yields than minted. Thb, I would prefer staying separate functions and merge only if it makes sense

mijovic: But we can also use estimate + what you explained and it should work. But is not exact...

yuval-weiss: good point ser.

yuval-weiss: sounds good

contracts/stats/Stats.sol
```diff
@@ -141,6 +141,30 @@ contract Stats is ChainlinkTokenPairPriceFeed, Versioned {
         principals += tempusAMM.getExpectedReturnGivenIn(principals, true);
     }
 
+    /// Gets the estimated amount of Shares and Lp token amounts
+    /// @param tempusAMM Tempus AMM to use to swap TYS for TPS
+    /// @param amount Amount of BackingTokens or YieldBearingTokens that would be deposited
+    /// @param isBackingToken If true, @param amount is in BackingTokens, otherwise YieldBearingTokens
+    /// @return principals Amount of Principals that user could receive in this action
+    function estimatedDepositAndLeverage(
+        ITempusPool tempusPool,
```
StefanVitor, line 150: tempusPool in param comment.

mijovic:
Updated and rebased on top of the latest master.

StefanVitor (approved):
All good in my opinion.

PR #535, opened 2022-02-22
Add AmplificationData struct in AMM constructor

StefanVitor:


yuval-weiss:
Nice. https://github.com/tempus-finance/tempus-protocol/blob/master/scripts/deploy.pool.lido.ts should also be updated :)

contracts/amm/TempusAMM.sol
```diff
@@ -99,8 +99,7 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
         string memory name,
         string memory symbol,
         ITempusPool pool,
-        uint256 amplificationStartValue,
-        uint256 amplificationEndValue,
+        AmplificationData memory amplificationData,
```
yuval-weiss, line 102: hmm I think it's a little messy to use the `AmplificationData` struct as an argument here because `AmplificationData.startTime` is not used

StefanVitor: The reason for that is next - https://github.com/tempus-finance/tempus-protocol/pull/531#pullrequestreview-887669059. It's not possible to have all arguments that are need in constructor because of "CompilerError: Stack too deep when compiling inline assembly: Variable headStart is 1 slot(s) too deep inside the stack. Error HH600: Compilation failed", so this is one of the solutions (second solution is token addresses array).

yuval-weiss: hmmm yeah... imho token array is better than using this struct without actually using one of its properties... @mijovic thoughts?

mijovic: We should document this instead. Also, we could have an option to start amp update later down the line, like in future, so it might be a good option to have struct. If we decide going with struct, we need good docs explaining that if startTime is zero it will be unused. WDYT?

StefanVitor: Yes, I agree with @mijovic, only to document this well (that startTime zero is not used), and struct is a better option in that case.

yuval-weiss: but that's not the case. Even if `startTime` is non-zero it will not be used. It's ignored in any case...

yuval-weiss: I really don't think it's a good idea

PR #536, opened 2022-02-22
test: update Deposit unit tests to use the new depositAndCheck utils

RedFox20:


StefanVitor (approved):
All good in my opinion.

PR #538, opened 2022-02-22
fix: properly return a failure code if any of the unit tests are unsu…

RedFox20:
…ccessful, but still make sure to run all the tests

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -87,9 +87,11 @@ if [ "$ONLY_POOL" == "" ] && [ "$ONLY_TOKEN" == "" ]; then
 fi
 
 # wait for results and check for failure
-FAIL=0
+FAILS=0
+TOTAL=0
 for pid in "${PIDS[@]}"; do
-  wait $pid || let "FAIL+=1"
+  let "TOTAL+=1"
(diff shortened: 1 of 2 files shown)


run_tests.sh
```diff
@@ -74,17 +75,28 @@ for pool in $pools; do
...
+  wait $pid || let "FAIL+=1"
+done
+
 end_time=$(date +%s)
 elapsed=$(( end_time - start_time ))
 echo ""
 echo elapsed time: $(date -ud "@$elapsed" +'%M min %S sec')
+
+if [ "$FAIL" != "0" ]; then
+  exit $FAIL
```
yuval-weiss, line 101: We should probably just `exit 1` here....

yuval-weiss: some exit codes are reserved. 1 is general error

yuval-weiss: https://tldp.org/LDP/abs/html/exitcodes.html

RedFox20: Ok, I changed it to `1` and to report how many of the runners failed @yuval-weiss

mijovic (approved):

PR #543, opened 2022-03-02
Rename params to tempusPool

StefanVitor:


RedFox20 (approved):
This is a much better way to get approved 👍🏻 

Personally, I would prefer terse names such as `pool`, but this is okay as well.

PR #545, opened 2022-03-04
feature: precompute yield share address to always be higher

RedFox20:


contracts/TempusPool.sol
```diff
@@ -105,8 +106,30 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable, Versioned
 
         uint8 backingDecimals = _backingToken != address(0) ? IERC20Metadata(_backingToken).decimals() : 18;
         backingTokenONE = 10**backingDecimals;
+
         principalShare = new PrincipalShare(this, principalsData.name, principalsData.symbol, backingDecimals);
-        yieldShare = new YieldShare(this, yieldsData.name, yieldsData.symbol, backingDecimals);
+        yieldShare = deployYieldShare(yieldsData, backingDecimals);
```
mijovic, line 111: can we add `assert(address(yieldShare) > address(principalShare))`?

contracts/TempusPool.sol
```diff
@@ -105,8 +106,30 @@ abstract contract TempusPool is ITempusPool, ReentrancyGuard, Ownable, Versioned
...
+
+    function deployYieldShare(TokenData memory data, uint8 decimals) private returns (IPoolShare) {
+        bytes memory bytecode = bytes.concat(
+            type(YieldShare).creationCode,
+            abi.encode(address(this), data.name, data.symbol, decimals)
+        );
+        bytes32 bytecodeHash = keccak256(bytecode);
+
+        // we want Yields to have higher address than Principals
+        address minAddress = address(principalShare);
```
mijovic, line 122: Nice, we can now read immutables during construction time. This was introduced with some of the latest versions :)

contracts/TempusPool.sol
mijovic: Like this 
```suggestion
        yieldShare = deployYieldShare(yieldsData, backingDecimals);
        assert(address(yieldShare) > address(principalShare));
```

mijovic: Like this 
```suggestion
        yieldShare = deployYieldShare(yieldsData, backingDecimals);
        assert(address(yieldShare) > address(principalShare));
```

RedFox20: Hmm, yeah, I think this is needed, because there is no strict guarantee inside `deployYieldShare`.
Maybe I should put the assert in there instead?

RedFox20: Ok, I added this to `deployYieldShare()`: 
```ts
assert(deployedAddr == predicted);
```

mijovic (approved):

PR #548, opened 2022-03-07
Update stable math to use 0.8.*

mijovic:


contracts/amm/math/StableMath.sol
```diff
@@ -0,0 +1,480 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+pragma solidity ^0.8.0;
+
+import "./Math.sol";
+import "./FixedPoint.sol";
+
+contract StableMath {
```
mijovic, line 8: we can make it library as well

contracts/amm/math/StableMath.sol
```diff
@@ -269,24 +249,21 @@ contract StableMath {
...
-        uint256 amountInWithoutFee = newBalanceTokenIndex.sub(balances[tokenIndex]);
+        uint256 amountInWithoutFee = newBalanceTokenIndex - balances[tokenIndex];
 
         // First calculate the sum of all token balances, which will be used to calculate
         // the current weight of each token
-        uint256 sumBalances = 0;
-        for (uint256 i = 0; i < balances.length; i++) {
-            sumBalances = sumBalances.add(balances[i]);
-        }
+        uint256 sumBalances = sumElements(balances);
```
RedFox20, line 256: For future performance improvement, if we can change the interface, so that we know balances is always `uint256[2] memory`, then we can change this to `uint256 sumBalances = balances[0] + balances[1];`

RedFox20: Alternatively, we can try not using arrays for the tokens

contracts/amm/math/StableMath.sol
```diff
@@ -305,18 +282,15 @@ contract StableMath {
...
 
         // Calculate the weighted balance ratio without considering fees
         uint256[] memory balanceRatiosWithoutFee = new uint256[](amountsOut.length);
         uint256 invariantRatioWithoutFees = 0;
         for (uint256 i = 0; i < balances.length; i++) {
             uint256 currentWeight = balances[i].divUp(sumBalances);
-            balanceRatiosWithoutFee[i] = balances[i].sub(amountsOut[i]).divUp(balances[i]);
-            invariantRatioWithoutFees = invariantRatioWithoutFees.add(balanceRatiosWithoutFee[i].mulUp(currentWeight));
+            balanceRatiosWithoutFee[i] = (balances[i] - amountsOut[i]).divUp(balances[i]);
```
RedFox20, line 292: Here as well, there is a lot of ceremony being done, because code does not assume amountsOut.length, but if we know the length is always 2, we can remove the loop and reduce gas cost

RedFox20:
I went through all the changes line by line, and everything seemed OK so far.

contracts/amm/math/FixedPoint.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+pragma solidity ^0.8.0;
+
+library FixedPoint {
```
RedFox20, line 5: So these are cleaned up functions from balancer FixedPoint.

Should we try to move these into our math folder as a Fixed256x18 library? Right now we have Fixed256xVar only.

RedFox20 (approved):

contracts/amm/math/FixedPoint.sol
mijovic: This is ideally what I wanted to do, but will do all of these as separate PRs

contracts/amm/math/StableMath.sol
mijovic: I want to remove arrays, but as a last step, once we port amm itself

PR #549, opened 2022-03-10
vault: Initial version of LPVaultV1

axic:


contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,229 @@
...
+
+    uint8 internal immutable tokenDecimals;
+
+    IERC20 public immutable yieldBearingToken;
+    uint256 private immutable oneYBT;
+
+    ITempusPool public pool;
+    ITempusAMM public amm;
+    // TODO: remove dependency on stats
+    Stats public stats;
```
RedFox20, line 41: I think you can just copy paste the code from Stats here. Since we might be removing most of the Stats.sol code in the future anyway.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    function decimals() public view virtual override returns (uint8) {
+        return tokenDecimals;
+    }
+
+    function previewDeposit(uint256 amount) public view returns (uint256 shares) {
+        uint256 supply = totalSupply();
+        // TODO: rounding
+        return (supply == 0) ? amount : amount.mulfV(supply, totalAssets());
```
RedFox20, line 66: This is a cool optimization. Maybe we should add a comment for the full formula so it's more clear?

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,229 @@
...
+        require(isAppropriateAMM(newPool, newAMM), "AMM is not for the correct Pool");
+
+        ITempusController controller = ITempusController(pool.controller());
+
+        // Redeem all LP tokens
+        uint256 maxLpTokensToRedeem = IERC20(address(amm)).balanceOf(address(this));
+        controller.exitTempusAMM(amm, pool, maxLpTokensToRedeem, 1, 1, false);
+
+        // Withdraw from the Pool
+        uint256 principals = IERC20(address(pool.principalShare())).balanceOf(address(this));
```
RedFox20, line 123: Really wish we can upgrade pool shares to inherit from `IERC20` already...

contracts/vault/LPVaultV1.sol
axic: By the time this PR is merged, I hope we remove the dependency.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,229 @@
...
+        require(amount != 0, "No shares would be burned");
+
+        _burn(msg.sender, shares);
+
+        // TODO: exitAmmGivenAmountsOutAndEarlyRedeem + exitAmmGivenLpAndRedeem
+        assert(false);
+    }
+
+    /// @return true if the given _amm uses the shares of the given _pool
+    function isAppropriateAMM(ITempusPool _pool, ITempusAMM _amm) private view returns (bool) {
```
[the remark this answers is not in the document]
axic: > This should be named isTempusPoolAMM

Why?

> commented as @return true if given TempusAMM uses shares of the given TempusPool.

Why?

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,253 @@
...
+            uint256 lpTokens
+        )
+    {
+        (principalsStaked, yieldsStaked) = amm.compositionBalanceOf(address(this));
+        lpTokens = IERC20(address(amm)).balanceOf(address(this));
+        principals = IERC20(address(pool.principalShare())).balanceOf(address(this));
+        yields = IERC20(address(pool.yieldShare())).balanceOf(address(this));
+
+        if (requiredShares > principals) {
+            principalsStaked = min(requiredShares - principals, principalsStaked);
```
axic, line 154: The `min` here and below may end up with less shares than needed, however we have the `require` statement to ensure these allocations add up correctly.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,253 @@
...
+        } else {
+            uint256 requiredShares = pool.getSharesAmountForExactTokensOut(amount, false);
+
+            (
+                uint256 principals,
+                uint256 yields,
+                uint256 principalsStaked,
+                uint256 yieldsStaked,
+                uint256 maxLpTokensToRedeem
+            ) = calculateWithdrawalShareSplit(requiredShares);
```
axic, line 115: Actually if not using the controller's helper, but directly implementing it here, this function wouldn't be needed.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,253 @@
...
+                principalsStaked,
+                yieldsStaked,
+                maxLpTokensToRedeem,
+                false
+            );
+        }
+    }
+
+    // FIXME: move to some generic helper file
+    function min(uint256 a, uint256 b) private pure returns (uint256 c) {
```
RedFox20, line 131: We have `contracts/math/` so adding a `Math.sol` there would make senses

contracts/vault/LPVaultV1.sol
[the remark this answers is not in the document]
RedFox20: For AMM's which deal with TempusPool shares, we already use terminology `TempusPoolAMM`. Makes sense?
As for the devdocs, instead of using parameter names directly using the conceptual names is immune to variable renaming

contracts/vault/LPVaultV1.sol
RedFox20: I looked at `exitAmmGivenAmountsOutAndEarlyRedeem` and it's quite involved. I don't know if it makes sense to duplicate that logic here

contracts/vault/LPVaultV1.sol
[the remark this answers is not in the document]
axic: > And implementation check could just be return `_amm.token0() == _pool.principalShare();` assuming that this will only be used for the newer pools.

It would still need to check `token0 == principal && token1 == yield`

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,253 @@
...
+    function previewDeposit(uint256 amount) public view returns (uint256 shares) {
+        uint256 supply = totalSupply();
+        // TODO: rounding
+        return (supply == 0) ? amount : amount.mulfV(supply, totalAssets());
+    }
+
+    function previewWithdraw(uint256 shares) public view returns (uint256 amount) {
+        uint256 supply = totalSupply();
+        // TODO: rounding
+        return (supply == 0) ? shares : shares.mulfV(totalAssets(), supply);
```
mijovic, line 72: What if we do actual withdrawals in a similar way how it is done in Pools/AMM. So, instead of calculating like this I would preview withdraw from TempusPool which will give exact amounts instead of approximation which is purely based on totalAssets calculation...

contracts/vault/LPVaultV1.sol
RedFox20: If it looks better, then it is worth giving a try

contracts/vault/LPVaultV1.sol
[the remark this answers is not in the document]
RedFox20: Hmm, thats true, we might get an AMM with token0 = Principals and token1 = faketoken, so indeed, can't skip on checking both.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,253 @@
...
+
+        // NOTE: at this point any leftover shares will be "lost"
+        // FIXME: decide what to do with leftover lp/principal/yield shares
+        pool = newPool;
+        amm = newAMM;
+        stats = newStats;
+    }
+
+    /// @return tokenAmount the current total balance in terms of YBT held by the vault
+    function totalAssets() private view returns (uint256 tokenAmount) {
```
mijovic, line 230: The idea I proposed in a call was to calculate `pricePerShare` instead of total asset. It would represent value of one LPVault token, and we can calculate it as follows:
```
uint256 ybtBalance = pool.yieldBearingToken.balanceOf(address(this));
uint256 lpTokens = IERC20(address(amm)).balanceOf(address(this));
uint256 principals = IERC20(address(pool.principalShare())).balanceOf(address(this));
uint256 yields = IERC20(address(pool.yieldShare())).balanceOf(address(this));

uint256 supply = totalSupply();
return ybtBalance.divfV(supply, ONE) + stats.estimateExitAndRedeem(
    amm,
    pool,
    lpTokens.divfV(supply, ONE),
    principals.divfV(supply, ONE),
    yields.divfV(supply, ONE),
    /*threshold*/
    0,
    false
);
```

axic: This is not much different from the estimation of a small percentage we discussed. You still get `totalAssets()` based on `pricePerShare() * totalSupply()`.

test/LPVault.test.ts
```diff
@@ -0,0 +1,80 @@
...
+  const ONE_HOUR:number = 60*60;
+  const ONE_DAY:number = ONE_HOUR*24;
+  const ONE_MONTH:number = ONE_DAY*30;
+  const ONE_YEAR:number = ONE_MONTH*12;
+  const ONE_AMP_UPDATE_TIME:number = ONE_DAY;
+
+  let stats:Stats;
+  let lpVault:LPVault;
+  
+  async function createPools(params:CreateParams): Promise<{pool:TempusPool, amm:TempusAMM}> {
```
axic, line 38: This is a copy of `test/amm/TempusAMM.test.ts`.

test/utils/LPVault.ts
```diff
@@ -0,0 +1,30 @@
...
+    super(contractName, decimals/*default decimals*/, contract);
+  }
+
+  static async create(pool: TempusPool, amm: TempusAMM, stats: Stats, name: string, symbol: string): Promise<LPVault> {
+    const lpVault = await ContractBase.deployContract("LPVaultV1", pool.address, amm.address, stats.address, name, symbol);
+    return new LPVault("LPVaultV1", await lpVault.decimals(), lpVault);
+  }
+
+  async deposit(amount: NumberOrString, recipient: SignerOrAddress): Promise<NumberOrString> {
+    return this.contract.deposit(amount, addressOf(recipient));
```
RedFox20, line 20: All the dynamic ethersjs calls return a `BigNumber` from `uint256` return values. In order to make it suitable for unit tests, it needs to be converted to a decimal like this:
```ts
return this.fromBigNum(await this.contract.deposit(amount, addressOf(recipient)));
```

test/LPVault.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  async function createVault(pool: TempusPool, amm: TempusAMM): Promise<void> {
+    stats = await Stats.create();
+    lpVault = await LPVault.create(pool, amm, stats, "Tempus LP Vault", "PVALT");
+  }
+
+  it("Smoke test", async () => {
+    const pool = await createPools({yieldEst:0.1, duration:ONE_MONTH, amplifyStart:5, amplifyEnd:5, ammBalancePrincipal: 10000, ammBalanceYield: 100000});
+    await createVault(pool.pool, pool.amm);
```
RedFox20, line 70: Probably it would be nicer for `createPools` to be something like `createPoolsAndVault`, since you will probably need it every time

test/utils/LPVault.ts
axic: Done.

test/LPVault.test.ts
axic: No, that's the point. Need to create new pool for migration.

RedFox20: Makes sense if that is needed. Can you rename it so it is more obvious somehow? When reading this test, I don't see anything explaining that a migration is happening

axic: The test called `migration` is doing a migration.

axic: > Probably it would be nicer for `createPools` to be something like `createPoolsAndVault`, since you will probably need it every time

@RedFox20 could you help with creating the helper?

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,309 @@
...
+        // FIXME: decide what to do with leftover lp/principal/yield shares
+        // Remove unlimited approval
+        yieldBearingToken.safeApprove(pool.controller(), 0);
+        amm.approve(pool.controller(), 0);
+        pool = newPool;
+        amm = newAMM;
+        stats = newStats;
+    }
+
+    function shutdown() external onlyOwner {
```
axic, line 277: Need test case for shutdown.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,309 @@
...
+        require(supply != 0, "PricePerShare for 0 supply is not allowed");
+
+        (rate, , , , ) = stats.estimateExitAndRedeem(
+            amm,
+            pool,
+            lpTokens.divfV(supply, oneYBT),
+            principals.divfV(supply, oneYBT),
+            yields.divfV(supply, oneYBT),
+            /*threshold*/
+            10 * onePoolShare,
```
axic, line 304: What threshold do we want? Stick with 0? Some small fraction?

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,301 @@
...
+    // poolShares -- equals BT
+    // LP -- fixed 18
+    // vaultShare -- equals YBT
+    constructor(
+        ITempusPool _pool,
+        ITempusAMM _amm,
+        Stats _stats,
+        string memory name,
+        string memory symbol
+    ) ERC20OwnerMintableToken(name, symbol) {
```
yuval-weiss, line 59: why `ERC20OwnerMintableToken` and not `ERC20`?

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,301 @@
...
+        require(isTempusPoolAMM(_pool, _amm), "AMM is not for the correct Pool");
```
yuval-weiss, line 60: can be done later when the PR is closer to being finalized but just noting it here - require statements should be replaced with custom errors

contracts/vault/LPVaultV1.sol
axic: Good point, it was just a remnant from old code, will change.

contracts/vault/LPVaultV1.sol
mijovic: We can also move `exitAmmGivenAmountsOutAndEarlyRedeem` here, as it is not used anywhere else. Also, it is related to LPing, so to me it makes sense to have it here instead

contracts/vault/LPVaultV1.sol
mijovic: what if we parametrize it. Then it will be up to the user/frontend to supply it every time?

contracts/vault/ILPVaultV1.sol
```diff
@@ -0,0 +1,43 @@
...
+import "../stats/Stats.sol";
+
+interface ILPVaultV1 is IERC20Metadata, IOwnable {
+    function yieldBearingToken() external view returns (IERC20);
+    
+    function pool() external view returns (ITempusPool);
+    function amm() external view returns (ITempusAMM);
+    function stats() external view returns (Stats);
+    
+    /// True if the vault is shut down. This mean depoists are disabled, but withdrawals
```
pashov, line 18: Sentence in doc not finished

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+        // FIXME: validate newStats too
+
+        uint256 preExit = yieldBearingToken.balanceOf(address(this));
+
+        // Withdraw from current pool
+        exitPool();
+
+        uint256 amount = yieldBearingToken.balanceOf(address(this));
+
+console.log("migrate balances", preExit, amount);
```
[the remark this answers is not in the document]
pashov: import as well

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+
+        // Deposit all yield bearing tokens to new pool
+        // Unlimited approval.
+//        yieldBearingToken.safeApprove(newPool.controller(), type(uint256).max);
+        yieldBearingToken.approve(newPool.controller(), type(uint256).max);
+        newAMM.approve(newPool.controller(), type(uint256).max);
+        if (amount > 0) {
+//            ITempusController(newPool.controller()).depositAndProvideLiquidity(newAMM, newPool, amount, false);
```
pashov, line 263: Why is this commented out, is there a problem with this

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+            amm,
+            pool,
+            lpTokens.divfV(supply, oneYBT),
+            principals.divfV(supply, oneYBT),
+            yields.divfV(supply, oneYBT),
+            /*threshold*/
+            10 * onePoolShare,
+            false
+        );
+        rate += ybtBalance.divfV(supply, oneYBT);
```
pashov, line 306: Is it possible to get to a denial of service if `ybtBalance * oneYBT` reverts due to overflow?

test/LPVault.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    console.log(await testFixture.userState(pool.amm.address));
+    await createVault(pool.pool, pool.amm);
+    console.log(await testFixture.userState(owner));
+    expect(await lpVault.balanceOf(owner)).to.equal(0);
+    await lpVault.ybt.approve(owner, lpVault.address, 200);
+    await lpVault.deposit(owner, 100, owner);
+    console.log("after1 amm", await testFixture.userState(pool.amm.address));
+    console.log("after1 lpv", await testFixture.userState(lpVault.address));
+    console.log(+await lpVault.balanceOf(owner));
+    expect(+await lpVault.balanceOf(owner)).to.be.within(99.9999, 100.0002);
```
pashov, line 80: Why the `+` before `await`

contracts/vault/ILPVaultV1.sol
```diff
@@ -0,0 +1,43 @@
...
+    
+    function isShutdown() external view returns (bool);
+
+    function previewDeposit(uint256 amount) external view returns (uint256 shares);
```
RedFox20, line 22: These also need basic docs, it's just a lot of typing, but quite easy to do in 15mins

contracts/vault/ILPVaultV1.sol
```diff
@@ -0,0 +1,43 @@
...
+    /// Withdraws `shares` of LPVault tokens.
+    /// @return amount The number of yield bearing tokens acquired.
+    function withdraw(uint256 shares, address recipient) external returns (uint256 amount);
+
+    /// Migrates all funds from the current pool to the new pool.
+    function migrate(
+        ITempusPool newPool,
+        ITempusAMM newAMM,
+        Stats newStats
+    ) external;
```
RedFox20, line 38: This can probably fit a single line

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+import "../utils/UntrustedERC20.sol";
+import "../ITempusController.sol";
+import "../ITempusPool.sol";
+import "../amm/ITempusAMM.sol";
+import "../stats/Stats.sol";
+
+import "hardhat/console.sol";
+
+/// This LP Vault follows a very simple strategy.
+/// TODO: make this EIP4626 compatible
```
RedFox20, line 19: What was the plan with this? Should we do it?

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+///
+/// Support needed from AMM:
+/// 1. Calculate current holdings in terms of YBT
+/// 2. Buy holdings in exchange of given YBT
+///    - How much lpVaultShare to give from given YBT?
+///    - This is solved by depositAndProvideLiquidity
+/// 3. Sell holdings to get required amount of YBT
+///    - How much principal/yield/LP (and how much lpVaultShare that represents) to sell for given YBT?
+contract LPVaultV1 is ERC20, Ownable {
```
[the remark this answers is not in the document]
RedFox20: Indeed. This should be `is ILPVaultV1`.
I'm not sure why we're using V1, probably we discussed this as such before this task was started

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+    using SafeERC20 for IERC20;
+    using UntrustedERC20 for IERC20;
+    using Fixed256xVar for uint256;
+
+    // The decimals are the same as yield bearing token.
+    uint8 internal immutable tokenDecimals;
+
+    IERC20 public immutable yieldBearingToken;
+    uint256 private immutable oneYBT; // also equals to oneVaultShare
```
RedFox20, line 37: Nice. This is helpful 👍🏻

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+    Stats public stats;
+
+    bool public isShutdown;
+
+    // NOTE about decimals
+    // BT -- variable
+    // YBT -- variable
+    // poolShares -- equals BT
+    // LP -- fixed 18
+    // vaultShare -- equals YBT
```
RedFox20, line 53: Yep, this is so confusing at times.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+        ITempusAMM _amm,
+        Stats _stats,
+        string memory name,
+        string memory symbol
+    ) ERC20(name, symbol) {
+        require(isTempusPoolAMM(_pool, _amm), "AMM is not for the correct Pool");
+        pool = _pool;
+        amm = _amm;
+        stats = _stats;
+        yieldBearingToken = IERC20(pool.yieldBearingToken());
```
RedFox20, line 65: We should probably store `yieldBearingToken` as `IERC20Metadata public immutable yieldBearingToken;` like we do in other contracts

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+        tokenDecimals = IERC20Metadata(address(yieldBearingToken)).decimals();
```
RedFox20, line 66: And then
```ts
tokenDecimals = yieldBearingToken.decimals();
```

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+
+        oneYBT = 10**tokenDecimals;
+        assert(
+            IERC20Metadata(address(pool.principalShare())).decimals() ==
+                IERC20Metadata(address(pool.yieldShare())).decimals()
+        );
```
RedFox20, line 72: `IPoolShare` is now `IERC20Metadata`, so this should work also:
```ts
        assert(pool.principalShare().decimals() == pool.yieldShare().decimals());
```

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+
+    // This mirrors the decimals of the underlying yield bearing token.
+    function decimals() public view virtual override returns (uint8) {
+        return tokenDecimals;
+    }
+
+    function previewDeposit(uint256 amount) public view returns (uint256 shares) {
+        uint256 supply = totalSupply();
+        // TODO: rounding
+        return (supply == 0) ? amount : amount.mulfV(supply, totalAssets());
```
RedFox20, line 91: `mulfV` is essentially `mulDown` so we could add a `mulUp` overload. No idea how to name that though.

contracts/vault/LPVaultV1.sol
RedFox20: That's a good solution. @axic what do you think?

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+            uint256 yields,
+            uint256 principalsStaked,
+            uint256 yieldsStaked,
+            uint256 lpTokens
+        )
+    {
+        (principalsStaked, yieldsStaked) = amm.compositionBalanceOf(address(this));
+        lpTokens = IERC20(address(amm)).balanceOf(address(this));
+        principals = IERC20(address(pool.principalShare())).balanceOf(address(this));
+        yields = IERC20(address(pool.yieldShare())).balanceOf(address(this));
```
RedFox20, line 179: I think these casts are no longer required

contracts/vault/LPVaultV1.sol
RedFox20: Should probably write that down as a brief comment

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+        // Redeem all LP tokens
+        uint256 maxLpTokensToRedeem = IERC20(address(amm)).balanceOf(address(this));
+        // TODO: exitGivenLpIn also calls this
+        (uint256 principals, uint256 yields) = amm.getTokensOutGivenLPIn(maxLpTokensToRedeem);
+        if (maxLpTokensToRedeem > 0) {
+            amm.exitGivenLpIn(maxLpTokensToRedeem, principals, yields, address(this));
+        }
+
+        // Withdraw from the Pool
+        principals = IERC20(address(pool.principalShare())).balanceOf(address(this));
```
RedFox20, line 212: Casts no longer necessary

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    function totalAssets() public view returns (uint256 tokenAmount) {
+        return pricePerShare().mulfV(totalSupply(), oneYBT);
+    }
+
+    /// Price per share in YBT.
+    function pricePerShare() private view returns (uint256 rate) {
+        uint256 ybtBalance = yieldBearingToken.balanceOf(address(this));
+        uint256 lpTokens = IERC20(address(amm)).balanceOf(address(this));
```
RedFox20, line 289: Casts no longer needed

tempus-sdk/tempus/LPVault.ts
```diff
@@ -0,0 +1,49 @@
...
+    await this.connect(caller).deposit(this.ybt.toBigNum(amount), addressOf(recipient));
+  }
+
+  async previewWithdraw(caller: Signer, shares: Numberish): Promise<Numberish> {
+    return this.ybt.fromBigNum(await this.connect(caller).previewWithdraw(this.ybt.toBigNum(shares)));
+  }
+
+  async withdraw(caller: Signer, shares: Numberish, recipient: Addressable): Promise<void> {
+    await this.connect(caller).withdraw(this.toBigNum(shares), addressOf(recipient));
+  }
```
RedFox20, line 36: Inconsistent utility use here, in preview you are using `this.ybt.toBigNum(shares)` but in actual withdraw you are using `this.toBigNum(shares)`.

For clarity and to avoid potential mistakes, both of them should use the same contract function. Although in the docs it was stated that `YBT.decimals == Vault.decimals` so there shouldn't be a bug here, just clarity issue 🙂

tempus-sdk/tempus/LPVault.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  async migrate(caller: Signer, pool: TempusPool, amm: TempusAMM, stats: Stats): Promise<void> {
+    await this.connect(caller).migrate(pool.address, amm.address, stats.address);
+  }
+
+  async shutdown(caller: Signer): Promise<void> {
+    await this.connect(caller).sthudown();
```
RedFox20, line 43: Typo here, it should be `shutdown()`

contracts/vault/ILPVaultV1.sol
```diff
@@ -0,0 +1,43 @@
...
+
+    function shutdown() external;
+
+    function totalAssets() external view returns (uint256 tokenAmount);
```
RedFox20, line 42: Here it should be documented in what precision the totalAssets() will be

tempus-sdk/tempus/LPVault.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  async totalAssets(caller: Signer): Promise<Numberish> {
+    return this.ybt.fromBigNum(await this.connect(caller).totalAssets());
```
RedFox20, line 47: Is totalAssets() in YBT or in VaultShares? Just semantics

tempus-sdk/tempus/PoolTestFixture.ts
```diff
@@ -365,7 +365,7 @@ export abstract class PoolTestFixture {
   /**
    * @returns Balances state for a single user
    */
-  async userState(user:Signer): Promise<UserState> {
+  async userState(user:Addressable): Promise<UserState> {
```
RedFox20, line 368: Nice! 👍🏻

tempus-sdk/utils/Utils.ts
```diff
@@ -38,23 +38,23 @@ export async function evmMineInSingleBlock(fn: ()=>Promise<void>): Promise<void>
 /**
  * Increase current EVM time by seconds
  */
-export async function increaseTime(addSeconds: number) : Promise<void> {
+export async function increaseTime(addSeconds: number): Promise<void> {
   await ethers.provider.send("evm_increaseTime", [addSeconds]);
-  await ethers.provider.send("evm_mine", []);
+  await evmMine();
```
RedFox20, line 43: Good catch 👍🏻

test/LPVault.test.ts
```diff
@@ -0,0 +1,160 @@
+import { expect } from "chai";
+import { BigNumber, ethers } from "ethers";
```
RedFox20, line 2: We shouldn't need `BigNumber` import here. And if we really have to pass a bignum to the contract directly, we can use `DecimalUtils.ts` function `bn()`.

test/LPVault.test.ts
```diff
@@ -0,0 +1,160 @@
...
+      ammAmplifyEnd: params.amplifyStart /*NOTE: using Start value here to not trigger update yet */
+    });
+
+    const tempusAMM = testFixture.amm;
+    [owner, user, user1] = testFixture.signers;
+
+    await testFixture.setupAccounts(owner, [[owner, 100_000], [user, 100_000]]);
+
+    const depositAmount = 1_000_000;
+    await testFixture.deposit(owner, depositAmount);
```
RedFox20, line 51: Does the `owner` have enough backing tokens to deposit?

test/LPVault.test.ts
```diff
@@ -0,0 +1,160 @@
...
+      initialRate:1.0, poolDuration:params.duration, yieldEst:params.yieldEst,
+      ammSwapFee:SWAP_FEE_PERC,
+      ammAmplifyStart: params.amplifyStart,
+      ammAmplifyEnd: params.amplifyStart /*NOTE: using Start value here to not trigger update yet */
+    });
+
+    const tempusAMM = testFixture.amm;
+    [owner, user, user1] = testFixture.signers;
+
+    await testFixture.setupAccounts(owner, [[owner, 100_000], [user, 100_000]]);
```
RedFox20, line 48: This is one of the old testing functions which is a bit unreliable. It might cause issues for you.

It deposits the sum of the parameter values including a small fixed amount: `100 + 100_000 + 100_000` and will receive ~some~ amount of YBT... Whether it is 200_100 YBT (pegged asset) or maybe it is a completely different number.

Regardless, it then transfers 100k backing tokens to each user and the desired YBT amounts you specified with the 100_000 value.

Basically, it's a mess and probably you are better off doing it yourself (sorry for the trouble🐶)

test/LPVault.test.ts
RedFox20: It automatically calls `valueOf()` from the underlying type, effectively converting the `Numberish` into a `number` with a single character. We use it a lot in unit tests.

tempus-sdk/tempus/LPVault.ts
axic: Good catch.

test/LPVault.test.ts
axic: This is some dead code from before the refactorings. it has been tough rebasing this over all the different refactorings which took place.

RedFox20: Yeah, makes sense, we've been trying to finish everything up quite fast, so everything has changed

contracts/vault/LPVaultV1.sol
axic: It is only using `estimateExitAndRedeem` now, removed other dependencies.

contracts/vault/ILPVaultV1.sol
axic: Docs were in the contract, not in the interface, moved.

contracts/vault/ILPVaultV1.sol
axic: Done.

contracts/vault/LPVaultV1.sol
axic: Yes, but need to update OpenZeppelin as they have merged it now. I think we should do after this PR.

contracts/vault/ILPVaultV1.sol
axic: Tell that to the linter.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,277 @@
...
+        yieldBearingToken.safeTransferFrom(msg.sender, address(this), amount);
+        ITempusController(pool.controller()).depositAndProvideLiquidity(amm, pool, amount, false);
+
+        _mint(recipient, shares);
+    }
+
+    function withdraw(uint256 shares, address recipient) external returns (uint256 amount) {
+        bool matured = pool.matured();
+
+        if (matured) {
```
axic, line 100: Should have a flag to avoid calling `exitPool` on every mature withdrawal.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,309 @@
...
+        uint256 supply = totalSupply();
+        // TODO: rounding
+        return (supply == 0) ? shares : shares.mulfV(totalAssets(), supply);
+    }
+
+    /// Deposits `amount` of yield bearing tokens.
+    /// @return shares The number of shares acquired.
+    function deposit(uint256 amount, address recipient) external returns (uint256 shares) {
+        // Quick exit path.
+        require(!isShutdown, "Vault is shut down.");
```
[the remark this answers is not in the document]
axic: Done.

contracts/vault/LPVaultV1.sol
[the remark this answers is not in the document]
axic: I think the plan was multiple stages with various complexity, but we can just drop it, cc @mijovic.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+
+        isShutdown = true;
+    }
+
+    function totalAssets() public view returns (uint256 tokenAmount) {
+        return pricePerShare().mulfV(totalSupply(), oneYBT);
+    }
+
+    /// Price per share in YBT.
+    function pricePerShare() private view returns (uint256 rate) {
```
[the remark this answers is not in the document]
axic: Changed.

contracts/vault/LPVaultV1.sol
axic: Moved the comment to the interface, but let's merge it that way and do it afterwards.

tempus-sdk/tempus/LPVault.ts
axic: vault.decimals == ybt.decimals

contracts/vault/LPVaultV1.sol
```diff
@@ -30,6 +32,9 @@ contract LPVaultV1 is ILPVaultV1, ERC20, Ownable {
 
     bool public isShutdown;
 
+    mapping(address => uint256) public withdrawalRequests;
+    uint256 public totalWithdrawalRequest;
```
axic: This was an idea for staged withdrawals, removed from this PR.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,302 @@
...
+        yieldBearingToken = pool.yieldBearingToken();
+        tokenDecimals = yieldBearingToken.decimals();
+        oneYBT = 10**tokenDecimals;
+        assert(pool.principalShare().decimals() == pool.yieldShare().decimals());
+        onePoolShare = 10**pool.principalShare().decimals();
+        // NOTE: this should be 18 decimals in every case, but asserting has the same cost
+        oneLP = 10**amm.decimals();
+
+        // Unlimited approval.
+        // NOTE: cannot use safeApprove here
```
axic, line 65:
```
     Error: VM Exception while processing transaction: reverted with reason string 'SafeERC20: approve from non-zero to non-zero allowance'
```

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,311 @@
...
+    function isTempusPoolAMM(ITempusPool _pool, ITempusAMM _amm) private view returns (bool) {
+        IPoolShare token0 = _amm.token0();
+        IPoolShare token1 = _amm.token1();
+        IPoolShare principalShare = _pool.principalShare();
+        IPoolShare yieldShare = _pool.yieldShare();
+        return (token0 == principalShare && token1 == yieldShare) || (token0 == yieldShare && token1 == principalShare);
+    }
+
+    // This function sets up approvals and replaces references.
+    function finalizeMigration(
```
axic, line 222: @yuval-weiss added this helper, but it makes the constructor more risky.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+        exitPool();
+
+        uint256 amount = yieldBearingToken.balanceOf(address(this));
+
+console.log("migrate balances", preExit, amount);
+
+        // Deposit all yield bearing tokens to new pool
+        // Unlimited approval.
+//        yieldBearingToken.safeApprove(newPool.controller(), type(uint256).max);
+        yieldBearingToken.approve(newPool.controller(), type(uint256).max);
```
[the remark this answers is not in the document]
axic: Here: https://github.com/tempus-finance/tempus-protocol/pull/549#discussion_r894512605

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+
+        uint256 supply = totalSupply();
+        require(supply != 0, "PricePerShare for 0 supply is not allowed");
+
+        (rate, , , , ) = stats.estimateExitAndRedeem(
+            amm,
+            pool,
+            lpTokens.divfV(supply, oneYBT),
+            principals.divfV(supply, oneYBT),
+            yields.divfV(supply, oneYBT),
```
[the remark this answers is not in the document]
axic: It can be non zero after withdrawals before maturity.

contracts/vault/LPVaultV1.sol
```diff
@@ -98,10 +100,11 @@ contract LPVaultV1 is ILPVaultV1, ERC20, Ownable {
     function withdraw(uint256 shares, address recipient) external returns (uint256 amount) {
         bool matured = pool.matured();
 
-        if (matured) {
+        if (matured && !isExited) {
```
axic, line 103: Should this really only check `amm.balanceOf(this) != 0`?

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,316 @@
...
+        // Set up initial details.
+        finalizeSetup(_pool, _amm, _stats);
+    }
+
+    /// This mirrors the decimals of the underlying yield bearing token.
+    function decimals() public view virtual override(ERC20, IERC20Metadata) returns (uint8) {
+        return tokenDecimals;
+    }
+
+    function previewDeposit(uint256 amount) public view returns (uint256 shares) {
```
axic, line 68: Just don't get why it doesn't complain that these must be `override`.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,317 @@
...
+        if (pool.matured()) {
+            revert VaultHasNoActivePool();
+        }
+
+        shares = previewDeposit(amount);
+        if (shares == 0) {
+            revert NoSharesMinted();
+        }
+
+        yieldBearingToken.safeTransferFrom(msg.sender, address(this), amount);
```
axic, line 95: Do we want a reentrancy guard around functions using non amm/pool tokens?

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,320 @@
...
+        if (!isTempusPoolAMM(newPool, newAMM)) {
+            revert InvalidPoolAMM(newPool, newAMM);
+        }
+        // FIXME: validate newStats too
+
+        // Withdraw from current pool
+        exitPool();
+
+        // NOTE: at this point any leftover shares will be "lost"
+        // FIXME: decide what to do with leftover lp/principal/yield shares
```
axic, line 268: Probably should add a safeguard that if these remain above a certain threshold then migration fails, at least then repeatedly calling `exitPool()` maybe a solution to recover funds, but after `migrate()` finishes all hope is lost.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,320 @@
...
+            revert InvalidPoolAMM(_pool, _amm);
+        }
+
+        yieldBearingToken = _pool.yieldBearingToken();
+        tokenDecimals = yieldBearingToken.decimals();
+        oneYBT = 10**tokenDecimals;
+        onePoolShare = 10**_pool.principalShare().decimals();
+
+        // Set up initial details.
+        finalizeSetup(_pool, _amm, _stats);
```
mijovic, line 61: why not move this function just after `constructor`?

axic: It needs `yieldBearingToken`.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,320 @@
...
+            revert CannotMigrateToSamePool();
+        }
+        if (!pool.matured()) {
+            // Only allow migration after maturity to avoid withdrawal risks (loss and/or lockup due to liquidity) from pool.
+            revert CurrentPoolNotMaturedYet();
+        }
+        if (newPool.yieldBearingToken() != yieldBearingToken) {
+            revert PoolYieldBearingTokenMismatch();
+        }
+        if (10**newPool.principalShare().decimals() != onePoolShare) {
```
RedFox20, line 256: You could compare decimals() only because pow is expensive?

test/LPVault.test.ts
```diff
@@ -0,0 +1,283 @@
...
+  ammBalanceYield?: number;
+  ammBalancePrincipal?:number;
+}
+
+type PoolAndAmm = {
+  pool: TempusPool;
+  amm: TempusPoolAMM;
+}
+
+describeForEachPool.only("LPVault", (testFixture:PoolTestFixture) =>
```
RedFox20, line 27: Need to remove .only

RedFox20:
Is there anything remaining here or should we rebase and merge?

contracts/vault/LPVaultV1.sol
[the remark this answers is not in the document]
mijovic: I think we can drop it, probably as part of other PR

mijovic (approved):

mijovic (dismissed):

contracts/vault/LPVaultV1.sol
axic: Created a tracking issue.

contracts/vault/LPVaultV1.sol
axic: Created tracking issue.

contracts/vault/LPVaultV1.sol
axic: Created tracking issue.

contracts/vault/LPVaultV1.sol
```diff
@@ -0,0 +1,308 @@
...
+    function pricePerShare() private view returns (uint256 rate) {
+        uint256 ybtBalance = yieldBearingToken.balanceOf(address(this));
+        uint256 lpTokens = IERC20(address(amm)).balanceOf(address(this));
+        uint256 principals = IERC20(address(pool.principalShare())).balanceOf(address(this));
+        uint256 yields = IERC20(address(pool.yieldShare())).balanceOf(address(this));
+
+        uint256 supply = totalSupply();
+        require(supply != 0, "PricePerShare for 0 supply is not allowed");
+
+        (rate, , , , ) = stats.estimateExitAndRedeem(
```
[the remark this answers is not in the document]
axic: Created tracking issue.

contracts/vault/LPVaultV1.sol
axic: Created tracking issue.

contracts/vault/LPVaultV1.sol
axic: Created tracking issue.

PR #550, opened 2022-03-11
ci: Add Echidna smart contract fuzzing/property-based testing

StefanVitor:


StefanVitor:
The main question is about this decreased coverage because I didn't see tests for Echidna contracts anywhere (on any project that uses Echidna). Should I add test and for that contracts how coverage should be above 90+% or?

StefanVitor:
> The main question is about this decreased coverage because I didn't see tests for Echidna contracts anywhere (on any project that uses Echidna). Should I add test and for that contracts how coverage should be above 90+% or?

I add Echidna files to skip for code coverage in .solcover.js file, I think that is good solution, but write your opinion about this.

mijovic:
> > The main question is about this decreased coverage because I didn't see tests for Echidna contracts anywhere (on any project that uses Echidna). Should I add test and for that contracts how coverage should be above 90+% or?
> 
> I add Echidna files to skip for code coverage in .solcover.js file, I think that is good solution, but write your opinion about this.

I think this is good, we should not cover code that is used only for testing. There is no value in that.

.circleci/config.yml
```diff
@@ -184,6 +184,33 @@ jobs:
...
+          command: |
+            sudo apt update
+            sudo apt install python3-pip
+            sudo pip3 install slither-analyzer
+            sudo pip3 install solc-select
+            sudo solc-select install 0.8.10
+            sudo solc-select use 0.8.10
+            sudo wget -O /tmp/echidna-test.tar.gz https://github.com/crytic/echidna/releases/download/v2.0.0/echidna-test-2.0.0-Ubuntu-18.04.tar.gz
+            sudo tar -xf /tmp/echidna-test.tar.gz -C /usr/bin
+            sudo chmod +x /usr/bin/echidna-test
```
mijovic, line 206: Is it running install every time we run this job?

StefanVitor: Yes, maybe to try to optimize that?

mijovic: I am not sure how we can do it, probably there is a way with docker containers or something like that? Let's maybe merge this PR without optimization and then deal with that later?

contracts/echidna/EchidnaTempusPool.sol
```diff
@@ -0,0 +1,157 @@
...
+            require(address(backingToken) == address(0), "given TempusPool's Backing Token is not ETH");
+        }
+
+        address recipient = address(uint160(recipientIndex % _ADDRESSES_LENGTH));
+        tempusPool.onDepositBacking{value: msg.value}(backingTokenAmount, recipient);
+    }
+
+    function _depositYieldBearing(uint256 yieldTokenAmount, uint256 recipientIndex) internal {
+        IERC20 yieldBearingToken = IERC20(tempusPool.yieldBearingToken());
+        yieldBearingToken.approve(address(this), yieldTokenAmount);
```
mijovic, line 120: Same here

mijovic: this is sort of contract approving itself to spend tokens, I think it is redundant

contracts/echidna/EchidnaTempusPool.sol
```diff
@@ -0,0 +1,157 @@
...
+        );
+
+        address recipient = address(uint160(recipientIndex % _ADDRESSES_LENGTH));
+        tempusPool.onDepositYieldBearing(transferredYBT, recipient);
+    }
+
+    /// @dev Check are principalShare total supply and yieldShare total supply
+    ///     equal before maturity
+    function echidnaPrincipalAndYieldEquality() public view returns (bool) {
+        if (tempusPool.matured() == false) {
```
mijovic, line 135:
```suggestion
        if (!tempusPool.matured()) {
```

contracts/echidna/EchidnaTempusPool.sol
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /// @dev Check if pool has enough yield bearing tokens
+    function echidnaPoolHasEnoughYBT() public returns (bool) {
+        uint256 principalAmount = IERC20(address(tempusPool.principalShare())).totalSupply();
+        uint256 yieldAmount = IERC20(address(tempusPool.yieldShare())).totalSupply();
+        uint256 yieldBearingTokenAmount = IERC20(tempusPool.yieldBearingToken()).balanceOf(address(tempusPool));
+        uint256 estimateYieldBearingTokenAmount = tempusPool.estimatedRedeem(principalAmount, yieldAmount, false);
+
+        tempusPool.updateInterestRate();
```
mijovic, line 149: we need to call this first thing in the method to update rates

mijovic:
Only a few things/questions, otherwise it looks good.
Also, will need to rebase to the latest master  + squash commits

.circleci/config.yml
yuval-weiss: instead of `node`

contracts/echidna/EchidnaTempusPool.sol
```diff
@@ -0,0 +1,157 @@
...
+                earlyRedeemPercent: earlyRedeemPercent,
+                matureRedeemPercent: matureRedeemPercent
+            })
+        );
+    }
+
+    /// @dev Transfers accumulated Yield Bearing Token (YBT) fees
+    ///     from this pool contract to account with index `recipientIndex`.
+    /// @param recipientIndex Index for address which will receive the specified amount of YBT
+    function transferFees(uint256 recipientIndex) public {
```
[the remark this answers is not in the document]
StefanVitor: Yes, on this place, we could use address. But on deposit and redeem functions, it's better to use a limited number of addresses (my idea is with index and % _ADDRESSES_LENGTH) because it will be very hard to find an address that success deposit, and after that with that address to success redeem.

.circleci/config.yml
RedFox20: Should probably do as @yuval-weiss suggested, just use echidna image instead of `node/default`

contracts/echidna/EchidnaTempusPool.sol
```diff
@@ -0,0 +1,157 @@
...
+    /// @param principalAmount Amount of Tempus Principals to redeem in PrincipalShare decimal precision
+    /// @param yieldAmount Amount of Tempus Yields to redeem in YieldShare decimal precision
+    /// @param recipientIndex Index for address of user that will receive yield bearing tokens
+    function redeemToYieldBearing(
+        uint256 fromIndex,
+        uint256 principalAmount,
+        uint256 yieldAmount,
+        uint256 recipientIndex
+    ) public {
+        address from = address(uint160(fromIndex % _ADDRESSES_LENGTH));
```
RedFox20, line 73: Is this `address(uint160(fromIndex % _ADDRESSES_LENGTH))` something really specific to this test case?
Seems like it should be turned into some utility function
```ts
/// @dev This function's purpose is XXX because of YYY ...
function addressFromIndex(uint256 index) internal pure returns(address) {
    return address(uint160(index % _ADDRESSES_LENGTH));
}
```

RedFox20: Otherwise, this code needs some sort of explanation why  we're using `fromIndex` and `recipientIndex` instead of `address from` and `address recipient`. I'm not familiar with echidna's limitations, so maybe you can explain it with devdocs

contracts/echidna/EchidnaTempusPool.sol
```diff
@@ -0,0 +1,157 @@
...
+
+        tempusPool.redeemToBacking(from, principalAmount, yieldAmount, recipient);
+    }
+
+    function _depositBacking(uint256 backingTokenAmount, uint256 recipientIndex) internal {
+        IERC20 backingToken = IERC20(tempusPool.backingToken());
+        if (msg.value == 0) {
+            require(address(backingToken) != address(0), "Pool requires ETH deposits");
+
+            backingToken.approve(address(this), backingTokenAmount);
```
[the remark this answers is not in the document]
StefanVitor: No, it needs approval (got error "ERC20: transfer amount exceeds allowance" and also, Echidna doesn't reach lines after calling untrustedTransferFrom)

contracts/echidna/EchidnaTempusPool.sol
StefanVitor: Same as previous -it needs approval (got error "ERC20: transfer amount exceeds allowance" and also, Echidna doesn't reach lines after calling untrustedTransferFrom)

.circleci/config.yml
mijovic: Nice one sers, good catch

contracts/echidna/EchidnaTempusPool.sol
mijovic: Probably it is doing some weird things internally. But also, it probably makes sense, otherwise fuzzer would need to call approvals.
However, would be nice to understand what exactly happens here

contracts/echidna/EchidnaTempusPool.sol
StefanVitor: > Otherwise, this code needs some sort of explanation why we're using `fromIndex` and `recipientIndex` instead of `address from` and `address recipient`. I'm not familiar with echidna's limitations, so maybe you can explain it with devdocs

Well, Echidna could generate and work with addresses, but in my opinion, it has a problem with random address generation, so a few times I got that in 100 000 generations didn't find the correct sequence that deposits and after that could redeem. So, my idea was to have a limited number of addresses that work with deposit and redeem, and in that way, it could be better tested. And the easiest way to do that is on this way - get a random number, a mod that number with address size that we want, and that number cast to address (so, if _ADDRESSES_LENGTH = 10 then addresses will be from 0x000000...000 to 0x000000...009). I will add function addressFromIndex and explain that function in a way that I write here, so I think that could be enough.

contracts/echidna/EchidnaTempusPool.sol
StefanVitor: The problem was that `untrustedTransferFrom` always require allowance (even on self), so I changed from `untrustedTransferFrom` to `untrustedTransfer` and removed `approve` (I think that I had some problems in TempusPool functions without this approval, but as I saw now, everything works correct, so I'll remove this approve and change to `untrustedTransfer`).

contracts/echidna/EchidnaTempusPool.sol
```diff
@@ -0,0 +1,157 @@
...
+        uint256 yieldBearingTokenAmount = IERC20(tempusPool.yieldBearingToken()).balanceOf(address(tempusPool));
+        uint256 estimateYieldBearingTokenAmount = tempusPool.estimatedRedeem(principalAmount, yieldAmount, false);
+
+        tempusPool.updateInterestRate();
+
+        if (yieldBearingTokenAmount < estimateYieldBearingTokenAmount) {
+            return false;
+        }
+
+        return true;
```
RedFox20, line 155:
```suggestion
        return estimateYieldBearingTokenAmount >= yieldBearingTokenAmount ;
```

contracts/echidna/EchidnaTempusPool.sol
```diff
@@ -0,0 +1,157 @@
...
+
+    uint256 private constant _ADDRESSES_LENGTH = 10;
+
+    uint256 internal constant MATURITY_TIME = 4 weeks;
+    uint256 internal constant EST_YIELD = 0.1 ether;
+    uint256 internal constant DEPOSIT_PERCENT = 0.5 ether;
+    uint256 internal constant EARLY_REDEEM_PERCENT = 1 ether;
+    uint256 internal constant MATURE_REDEEM_PERCENT = 0.5 ether;
+
+    TempusPool internal tempusPool;
```
yuval-weiss, line 18: Can we inherit from the actual `XXXTempusPool` contracts instead of holding a reference?

yuval-weiss:
Are we gonna add `TempusController` too?

.solcover.js
```diff
@@ -2,7 +2,10 @@ module.exports = {
   skipFiles: [
     'mocks/lido/StETH.sol', // 1:1 with Lido, already tested by upstream
     'mocks/aave/WadRayMath.sol', // 1:1 with Aave, already tested by upstream
-    'amm/TempusAMMFactory.sol' // Balancer implementation
+    'amm/TempusAMMFactory.sol', // Balancer implementation
+    'echidna/EchidnaLidoTempusPool.sol', // Echidna contract
+    'echidna/EchidnaTempusPool.sol', // Echidna contract
+    'echidna/EchidnaYearnTempusPool.sol' // Echidna contract
```
yuval-weiss, line 8: Are we gonna add `TempusController` too?

mijovic: we need good properties to check with echidna, but the plan is to increase the amount of code we cover with fuzzing.
We should add it as separate PRs imo

contracts/echidna/EchidnaTempusPool.sol
mijovic: Sounds good to me

contracts/echidna/EchidnaTempusPool.sol
StefanVitor: This should be change to `return yieldBearingTokenAmount >= estimateYieldBearingTokenAmount;`, not to `return estimateYieldBearingTokenAmount >= yieldBearingTokenAmount;`?

contracts/echidna/EchidnaTempusPool.sol
StefanVitor: In this case, no, the reason is because the constructor that is called by Echidna must be empty, and there is no way to create and pass `lidoMock` address for LidoTempusPool or `asset` and `yearnVaultMock` address to YearnTempusPool (so, I must create that references directly in the constructor and after that to pass to TempusPool constructor).

.solcover.js
StefanVitor: Yes, as @mijovic wrote, we are not using TempusController for now, so that's the reason because it's not added here. But in the future, when we cover more properties, we will add TempusController.

contracts/echidna/EchidnaTempusPool.sol
yuval-weiss: Hmmm... What if we do inherit from the actual TempusPools (for example, `EchidnaYearnTempusPool is YearnTempusPool`), and then hardcode the constructor arguments in for `YearnTempusPool` in `EchidnaYearnTempusPool`. We can precalculate the addresses for `asset`/`lidoMock` or whatever if we deploy these using `create2`... Would that work?

.circleci/config.yml
StefanVitor: Could someone help me with this, I was trying different ways to use the echidna docker image and run the echidna-test, but in every way, there is some problem (I was trying and local and here)? @RedFox20 @yuval-weiss

contracts/echidna/EchidnaTempusPool.sol
StefanVitor: I'm not sure I understand where we could precalculate the addresses for `asset/lidoMock`. Could you provide a code example for that (maybe in that way I'll better understand)?

StefanVitor: For now, as we talked, I didn't find an appropriate solution, because if we inherit YearnTempusPool in one contract and LidoTempusPool in another contract, we will have a problem using EchidnaTempusPool because, in most functions of that abstract contract, it needs to have reference to tempusPool (and with this changes, we will not have `tempusPool` state variable). Also, with that inherits, it appears some new problems. Every solution that I try is more complicated and worse than this. Maybe we could try to refactor this in the future, but I think this should stay for now.

mijovic (approved):

PR #551, opened 2022-03-14
amm: Add compositionBalanceOf() helper

axic:


contracts/amm/TempusAMM.sol
```diff
@@ -287,6 +287,20 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
                 );
     }
 
+    function compositionBalanceOf(address account)
```
RedFox20, line 290: Needs devdocs to explain what's up with this :)

contracts/amm/TempusAMM.sol
```diff
@@ -287,6 +287,20 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
+        external
+        view
+        returns (uint256 token0balance, uint256 token1balance)
+    {
+        (, uint256[] memory balances, ) = getVault().getPoolTokens(getPoolId());
+        assert(balances.length == 2);
```
RedFox20, line 296: Is this assert needed, given that we know there are only 2 tokens anyway?

RedFox20:
This looks good, only needs documentation and then it's ready to merge

contracts/amm/TempusAMM.sol
axic: Yeah none of the others had a description, but this reminds me I forgot to put it into `ITempusAMM.sol`. Not that that file has any natspec either :)

contracts/amm/TempusAMM.sol
axic: I was wondering about the same, we don't check it elsewhere, but not sure if that is safe or not.

test/utils/TempusAMM.ts
```diff
@@ -105,6 +105,11 @@ export class TempusAMM extends ContractBase {
     return this.fromBigNum(await this.contract.balanceOf(user.address));
   }
 
+  async compositionBalanceOf(user:Signer): Promise<{token0: NumberOrString, token1: NumberOrString}> {
```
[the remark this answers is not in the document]
axic: Sure. Most of this file doesn't have dedvocds :(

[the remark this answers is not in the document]
RedFox20: Deficiency probably from october deathmarch. We kind of have them in some typescript functions and not in other places. What do you think should be correct approach?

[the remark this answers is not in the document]
axic: I added the comment in this PR. Usual best practice is to enforce it adding on new functions whenever they are introduced, and then the rest can be filled out in spare cycles.

contracts/amm/TempusAMM.sol
RedFox20: I think it's fine not to check it, because nothing else does and the AMM is always set up with 2

contracts/amm/interfaces/ITempusAMM.sol
```diff
@@ -80,4 +80,10 @@ interface ITempusAMM {
     /// @dev This function returns the appreciation of one BPT relative to the
     /// underlying tokens. This starts at 1 when the pool is created and grows over time
     function getRate() external view returns (uint256);
+
+    /// @dev Returns the balances the given account controlers in terms of token0/token1
+    /// via the LP tokens they have.
+    /// @return token0Balance Amount of Token0 corresponding to the LP tokens
+    /// @return token1Balance Amount of Token1 corresponding to the LP tokens
+    function compositionBalanceOf(address account) external view returns (uint256 token0Balance, uint256 token1Balance);
```
mijovic, line 88: Needs docs for account param

mijovic (commented):
Only this https://github.com/tempus-finance/tempus-protocol/pull/551/files#r827739632
Otherwise we can merge it

contracts/amm/interfaces/ITempusAMM.sol
axic: Done.

axic:
@mijovic updated.

mijovic (approved):

PR #554, opened 2022-03-15
Non fungible position manager

yuval-weiss:


contracts/PositionManager.sol
```diff
@@ -0,0 +1,111 @@
...
+
+        uint256 tokenAmount;
+        if (depositedAsset != address(0)) {
+            tokenAmount = IERC20(depositedAsset).untrustedTransferFrom(msg.sender, address(this), params.tokenAmount);
+        } else {
+            tokenAmount = msg.value;
+        }
+
+        (uint256 capitalsMinted, uint256 yieldsMinted) = ITempusController(params.tempusPool.controller())
+            .depositAndLeverage{value: msg.value}(
```
mijovic, line 61: Maybe this is better format? the one that we have atm looks pretty weird (depositAndLeverage looks like a param)
```suggestion
        (
            uint256 capitalsMinted, 
            uint256 yieldsMinted
        ) = ITempusController(params.tempusPool.controller()).depositAndLeverage{value: msg.value}(
```

mijovic:
This looks like a good direction to me. What it needs is tests

contracts/PositionManager.sol
```diff
@@ -0,0 +1,120 @@
...
+
+    constructor(string memory name, string memory symbol) ERC721(name, symbol) {}
+
+    function mintLeveragedPosition(MintParams calldata params) external payable nonReentrant returns (uint256 tokenId) {
+        address controller = params.tempusPool.controller();
+        address depositedAsset = params.isBackingToken
+            ? params.tempusPool.backingToken()
+            : params.tempusPool.yieldBearingToken();
+
+        uint256 tokenAmount;
```
StefanVitor, line 54: I'll write on this way
```suggestion
        uint256 tokenAmount = msg.value;
        if (depositedAsset != address(0)) {
            tokenAmount = IERC20(depositedAsset).untrustedTransferFrom(msg.sender, address(this), params.tokenAmount);
            IERC20(depositedAsset).approve(controller, tokenAmount);
        }
```

contracts/PositionManager.sol
```diff
@@ -0,0 +1,120 @@
...
+        uint256 deadline;
+        bool toBackingToken;
+    }
+
+    mapping(uint256 => Position) public positions;
+    uint256 private _nextId = 1;
+
+    constructor(string memory name, string memory symbol) ERC721(name, symbol) {}
+
+    function mintLeveragedPosition(MintParams calldata params) external payable nonReentrant returns (uint256 tokenId) {
```
StefanVitor, line 48: NatSpec

contracts/PositionManager.sol
```diff
@@ -0,0 +1,120 @@
...
+
+        positions[tokenId] = Position({
+            tempusAMM: params.tempusAMM,
+            tempusPool: params.tempusPool,
+            capitals: capitalsMinted,
+            yields: yieldsMinted
+        });
+    }
+
+    function burnPosition(BurnParams calldata params) external nonReentrant {
```
StefanVitor, line 84: NatSpec

contracts/PositionManager.sol
```diff
@@ -0,0 +1,120 @@
...
+        uint256 yieldsRate;
+        uint256 maxSlippage;
+        uint256 deadline;
+        bool toBackingToken;
+    }
+
+    mapping(uint256 => Position) public positions;
+    uint256 private _nextId = 1;
+
+    constructor(string memory name, string memory symbol) ERC721(name, symbol) {}
```
StefanVitor, line 46: Add Solhint disable
```suggestion
    // solhint-disable-next-line no-empty-blocks
    constructor(string memory name, string memory symbol) ERC721(name, symbol) {}
```

contracts/PositionManager.sol
yuval-weiss: done

contracts/PositionManager.sol
yuval-weiss: I changed a lot of things, lemme know if you still think it looks weird.

contracts/IPositionManager.sol
```diff
@@ -0,0 +1,52 @@
...
+    /// @return tokenId of the minted position
+    function mint(MintParams calldata params) external payable returns (uint256 tokenId);
+
+    /// @dev Burns a non-fungible positions and liquidates its value to the position owner
+    /// @param tokenId Token ID to burn
+    /// @param params Instructions for position liquidation
+    function burn(uint256 tokenId, BurnParams calldata params) external;
+
+    /// @dev retrieves position data for a given token ID.
+    function positions(uint256 tokenId) external view returns (Position memory);
```
axic, line 51: Perhaps `positionOf` is more in line with other functions?

contracts/PositionManager.sol
```diff
@@ -0,0 +1,110 @@
...
+            params.deadline
+        );
+
+        if (params.toBackingToken) {
+            IERC20(position.tempusPool.backingToken()).transfer(msg.sender, liquidatedTokenAmount);
+        } else {
+            IERC20(position.tempusPool.yieldBearingToken()).transfer(msg.sender, liquidatedTokenAmount);
+        }
+    }
+
```
axic, line 106: Ah, marking the variable public generates a getter returning all the fields and not a struct, right?

contracts/PositionManager.sol
```diff
@@ -0,0 +1,110 @@
...
+            params.yieldsRate,
+            params.maxSlippage,
+            params.toBackingToken,
+            params.deadline
+        );
+
+        if (params.toBackingToken) {
+            IERC20(position.tempusPool.backingToken()).transfer(msg.sender, liquidatedTokenAmount);
+        } else {
+            IERC20(position.tempusPool.yieldBearingToken()).transfer(msg.sender, liquidatedTokenAmount);
```
axic, line 103: These should use safetransfers.

contracts/PositionManager.sol
```diff
@@ -0,0 +1,110 @@
...
+                tokenAmount,
+                params.isBackingToken,
+                params.worstAcceptableCapitalsRate,
+                params.deadline
+            );
+        } else {
+            revert InvalidLeverageMultiplier();
+        }
+
+        _mint(params.recipient, (tokenId = _nextId++));
```
axic, line 63: Shouldn't this use `_safeMint`?

contracts/PositionManager.sol
```diff
@@ -0,0 +1,110 @@
...
+        _positions[tokenId] = Position({
+            tempusAMM: params.tempusAMM,
+            tempusPool: params.tempusPool,
+            capitals: capitalsMinted,
+            yields: yieldsMinted
+        });
+    }
+
+    function burn(uint256 tokenId, BurnParams calldata params) external override nonReentrant {
+        require(_isApprovedOrOwner(msg.sender, tokenId), "PositionManager: transfer caller is not owner nor approved");
```
axic, line 74: Should use custom error.

contracts/PositionManager.sol
```diff
@@ -0,0 +1,110 @@
...
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+
+import "./ITempusController.sol";
+import "./IPositionManager.sol";
+import "./utils/UntrustedERC20.sol";
+import "./math/Fixed256xVar.sol";
+
+contract PositionManager is IPositionManager, ERC721, ReentrancyGuard {
```
axic, line 14: Shouldn't this implement `ERC721Enumerable` so the list of positions is easily queryable? (Especially useful for the frontend.)

contracts/PositionManager.sol
```diff
@@ -0,0 +1,110 @@
...
+    function mint(MintParams calldata params) external payable override nonReentrant returns (uint256 tokenId) {
+        address controller = params.tempusPool.controller();
+        address depositedAsset = params.isBackingToken
+            ? params.tempusPool.backingToken()
+            : params.tempusPool.yieldBearingToken();
+
+        uint256 tokenAmount = msg.value;
+        if (depositedAsset != address(0)) {
+            tokenAmount = IERC20(depositedAsset).untrustedTransferFrom(msg.sender, address(this), params.tokenAmount);
+            IERC20(depositedAsset).approve(controller, tokenAmount);
```
axic, line 34: Shouldn't this use `safeIncreaseAllowance`?

contracts/IPositionManager.sol
yuval-weiss: I think `positionOf` is a bit confusing, because of the similarity to `balanceOf` I would expect it to accept a `address owner`

contracts/PositionManager.sol
yuval-weiss: done

contracts/PositionManager.sol
yuval-weiss: yup, thanks!

contracts/PositionManager.sol
yuval-weiss: Ja

contracts/PositionManager.sol
yuval-weiss: Actually the reason I didn't just use a public mapping is that it's not possible to override an interface-defined getter that returns a struct with a mapping in the implementation.

yuval-weiss: hmmm yeah makes sense

contracts/IPositionManager.sol
```diff
@@ -0,0 +1,52 @@
...
+        address recipient;
+        bool isBackingToken;
+    }
+
+    struct BurnParams {
+        uint256 maxLeftoverShares;
+        uint256 yieldsRate;
+        uint256 maxSlippage;
+        uint256 deadline;
+        bool toBackingToken;
```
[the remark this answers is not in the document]
yuval-weiss: yeah that's a nice feature. I added it.

contracts/IPositionManager.sol
```diff
@@ -0,0 +1,54 @@
...
+        address recipient;
```
axic, line 35: Nit: swap this to match the order of `MintParams`.

contracts/PositionManager.sol
```diff
@@ -0,0 +1,117 @@
...
+        if (!_isApprovedOrOwner(msg.sender, tokenId)) {
+            revert UnauthorizedBurn();
+        }
+
+        Position memory position = _positions[tokenId];
+
+        delete _positions[tokenId];
+        _burn(tokenId);
+
+        IERC20(address(position.tempusPool.principalShare())).safeIncreaseAllowance(
```
axic, line 84: These don't need `safeIncreaseAllowance` as we trust the shares.

contracts/PositionManager.sol
```diff
@@ -0,0 +1,117 @@
...
+                tokenAmount,
+                params.isBackingToken,
+                params.worstAcceptableCapitalsRate,
+                params.deadline
+            );
+        } else {
+            revert InvalidLeverageMultiplier();
+        }
+
+        _safeMint(params.recipient, (tokenId = _nextId++));
```
axic, line 64: I feel a bit uneasy about nested expressions, would prefer pulling out `tokenId = _nextId++;` to be a statement on its own.

contracts/PositionManager.sol
```diff
@@ -0,0 +1,117 @@
...
+import "./IPositionManager.sol";
+import "./utils/UntrustedERC20.sol";
+import "./math/Fixed256xVar.sol";
+
+contract PositionManager is IPositionManager, ERC721, ReentrancyGuard {
+    using Fixed256xVar for uint256;
+    using SafeERC20 for IERC20;
+    using UntrustedERC20 for IERC20;
+
+    mapping(uint256 => Position) internal _positions;
```
axic, line 20: Nit: `private`.

contracts/PositionManager.sol
mijovic: I agree with this. I feel pretty scary when I see this :)

contracts/IPositionManager.sol
```diff
@@ -0,0 +1,54 @@
...
+        ITempusAMM tempusAMM;
+        ITempusPool tempusPool;
+        uint128 capitals;
+        uint128 yields;
+    }
+
+    struct MintParams {
+        ITempusAMM tempusAMM;
+        ITempusPool tempusPool;
+        uint256 leverageMultiplier; /// for fixed positions, use leverageMultiplier=0
```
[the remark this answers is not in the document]
mijovic: I think 1e18 would be 1x leverage which means you have same amount of capitals and yields

contracts/PositionManager.sol
mijovic: +1

contracts/PositionManager.sol
```diff
@@ -0,0 +1,117 @@
...
+            position.tempusPool.controller(),
+            position.yields
+        );
+        uint256 liquidatedTokenAmount = ITempusController(position.tempusPool.controller()).exitAmmGivenLpAndRedeem(
+            position.tempusAMM,
+            position.tempusPool,
+            0,
+            position.capitals,
+            position.yields,
+            0,
```
mijovic, line 98: maybe adding comment for `0`s in this call. What they represent and why they are zeros would be great

contracts/IPositionManager.sol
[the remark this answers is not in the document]
axic: The leverage code only allows `>1e18`, so not sure if it supposed to allow 1x leverage then.

contracts/PositionManager.sol
yuval-weiss: done ser

contracts/PositionManager.sol
```diff
@@ -0,0 +1,142 @@
...
+    using UntrustedERC20 for IERC20;
+
+    mapping(uint256 => Position) private _positions;
+    uint256 private _nextId = 1;
+
+    // solhint-disable-next-line no-empty-blocks
+    constructor(string memory name, string memory symbol) ERC721(name, symbol) {}
+
+    function mint(MintParams calldata params) external payable override nonReentrant returns (uint256 tokenId) {
+        ITempusPool tempusPool = params.tempusAMM.token0().pool();
```
axic, line 27: Should we add a safety check here for `token0().pool() == token1().pool()` since all these are user submitted fields now?

contracts/PositionManager.sol
```diff
@@ -0,0 +1,142 @@
...
+            );
+        } else {
+            revert InvalidLeverageMultiplier();
+        }
+
+        tokenId = _nextId++;
+        _safeMint(params.recipient, tokenId);
+
+        _positions[tokenId] = Position({
+            capitals: SafeCast.toUint128(capitalsMinted),
```
axic, line 74: So this would fit 340282366920938463463.374607431768211455 under 18 decimals. *Should be safe (TM)*

contracts/PositionManager.sol
```diff
@@ -0,0 +1,142 @@
...
+        } else {
+            IERC20(tempusPool.yieldBearingToken()).safeTransfer(params.recipient, liquidatedTokenAmount);
+        }
+    }
+
+    function positions(uint256 tokenId) external view override returns (Position memory) {
+        return _positions[tokenId];
+    }
+
+    function exitAmmGivenLpAndRedeem(
```
axic, line 114: Hmm, is this code duplication?

contracts/PositionManager.sol
yuval-weiss: Yes, should be sifu.

contracts/IPositionManager.sol
axic: I'd still prefer something like `position` and not plural.

contracts/PositionManager.sol
axic: Well you could just use an immutable for pool/amm and deploy a position manager for each pool :)

yuval-weiss: Yeah but that would complicate the process of fetching all positions for a user... better to leave it as is imo

contracts/PositionManager.sol
yuval-weiss: I separated this block of logic from the main `burn` function because I was getting a `Stack too deep` compile error.

yuval-weiss: I renamed `exitAmmGivenLpAndRedeem` to `_liquidatePosition` though

contracts/PositionManager.sol
yuval-weiss: The TempusController verifies the provided `tempusAMM` is whitelisted, but I do think it's a good assertion to add to make the PositionManager a bit more resilient and safe from possible future changes such as removing the whitelist from the controller.

contracts/PositionManager.sol
```diff
@@ -0,0 +1,142 @@
...
+            position.yields,
+            params.maxLeftoverShares,
+            params.yieldsRate,
+            params.maxSlippage,
+            params.toBackingToken,
+            params.deadline
+        );
+
+        if (params.toBackingToken) {
+            IERC20(tempusPool.backingToken()).safeTransfer(params.recipient, liquidatedTokenAmount);
```
StefanVitor, line 104: I would use `transfer` because we trust shares (same on else part).

contracts/PositionManager.sol
axic: Ah ok, why the helper?

StefanVitor (approved):
From my side, all is good.

axic:
@yuval-weiss can you squash the commits here please?

contracts/IPositionManager.sol
```diff
@@ -0,0 +1,52 @@
...
+        uint128 capitals;
+        uint128 yields;
+        ITempusAMM tempusAMM;
+    }
+
+    struct MintParams {
+        ITempusAMM tempusAMM;
+        uint256 leverageMultiplier; /// for fixed positions, use leverageMultiplier=0
+        uint256 tokenAmount;
+        uint256 worstAcceptableCapitalsRate;
```
axic, line 21: Can you describe these fields? It seems this one works differently  depending on leverage.

contracts/IPositionManager.sol
```diff
@@ -0,0 +1,52 @@
...
+    struct Position {
+        uint128 capitals;
+        uint128 yields;
+        ITempusAMM tempusAMM;
+    }
+
+    struct MintParams {
+        ITempusAMM tempusAMM;
+        uint256 leverageMultiplier; /// for fixed positions, use leverageMultiplier=0
+        uint256 tokenAmount;
```
axic, line 20: Potentially would rename this to `depositedTokenAmount` or `tokenAmountToDeposit`. At least describe it here.

contracts/PositionManager.sol
yuval-weiss: It helped work around the `Stack too deep` error

contracts/PositionManager.sol
```diff
@@ -0,0 +1,143 @@
...
+        {
+            address depositedAsset = params.isBackingToken ? tempusPool.backingToken() : tempusPool.yieldBearingToken();
+
+            if (depositedAsset != address(0)) {
+                tokenAmountToDeposit = IERC20(depositedAsset).untrustedTransferFrom(
+                    msg.sender,
+                    address(this),
+                    params.tokenAmountToDeposit
+                );
+                IERC20(depositedAsset).approve(controller, tokenAmountToDeposit);
```
axic, line 41: Oh this should be `safeIncreaseApproval` as it is an untrusted token.

contracts/IPositionManager.sol
yuval-weiss: done ser

contracts/PositionManager.sol
yuval-weiss: done

contracts/PositionManager.sol
yuval-weiss: closing this

axic (changes requested):
Missing test cases for:
- mint with token(0), e.g. lido
- burn to backing token

I would also make sure there are all 4 cases done with lido and non-lido to have full coverage of weird token options.

mijovic (approved):

contracts/PositionManager.sol
```diff
@@ -0,0 +1,147 @@
...
+        ITempusAMM amm,
+        ITempusPool tempusPool,
+        uint128 capitals,
+        uint128 yields,
+        uint256 maxLeftoverShares,
+        uint256 yieldsRate,
+        uint256 maxSlippage,
+        bool toBackingToken,
+        uint256 deadline
+    ) private returns (uint256 liquidatedtokenAmount) {
```
axic, line 128:
```suggestion
    ) private returns (uint256 liquidatedTokenAmount) {
```

This?

yuval-weiss: yes :)

contracts/IPositionManager.sol
```diff
@@ -0,0 +1,53 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.10;
+
+import "./amm/interfaces/ITempusAMM.sol";
+import "./ITempusPool.sol";
+
+interface IPositionManager {
```
axic, line 7: I think this should inherit `IERC721, IERC721Metadata`.

contracts/IPositionManager.sol
```diff
@@ -0,0 +1,53 @@
...
+    error InvalidLeverageMultiplier(); /// Provided Leverage Multiplier is invalid
```
axic, line 8: Natspec should be above the line, not next to it.

contracts/IPositionManager.sol
```diff
@@ -0,0 +1,53 @@
...
+    error AmmSharesPoolMismatch(); /// A Tempus AMM whose shares are of different TempusPools was provided
+
+    struct Position {
+        uint128 capitals;
+        uint128 yields;
+        ITempusAMM tempusAMM;
+    }
+
+    struct MintParams {
+        ITempusAMM tempusAMM; /// the Tempus AMM (and its Tempus Pool) to use for position creation
```
axic, line 19: Does natspec work here above the struct members? Even if it is ignored there, would prefer there.

yuval-weiss: done

contracts/IPositionManager.sol
yuval-weiss: I agree

mijovic (approved):
To me, this looks ready to go. Everything is fixed.

contracts/PositionManager.sol
```diff
@@ -0,0 +1,147 @@
...
+        uint128 capitals,
+        uint128 yields,
+        uint256 maxLeftoverShares,
+        uint256 yieldsRate,
+        uint256 maxSlippage,
+        bool toBackingToken,
+        uint256 deadline
+    ) private returns (uint256 liquidatedTokenAmount) {
+        address controller = tempusPool.controller();
+        IERC20(address(tempusPool.principalShare())).approve(controller, capitals);
```
axic, line 130: I think the `IERC20` wrapping is not needed now, since ammv2 was merged.

RedFox20: Not yet because IPoolShare doesn't implement IERC20, but this is one of my unfinished PR-s which was waiting for amm/v2 merge.
I'll quickly put it up, so it can be merged to master and this PR can use it...

axic: Ah right due to the conflict of the `OwnerMintable`, I think I did that change locally too, a few times already. Let's hope we can merge it finally.

PR #555, opened 2022-03-15
AMMv2 remove input arrays from StableMath

RedFox20:


contracts/amm/math/StableMath.sol
```diff
@@ -62,11 +53,11 @@ library StableMath {
...
         uint256 ampTimesTotal = amplificationParameter * numTokens;
 
+        // TODO: Cleanup & simplify math
         for (uint256 i = 0; i < 255; i++) {
-            uint256 P_D = balances[0] * numTokens;
-            for (uint256 j = 1; j < numTokens; j++) {
-                P_D = Math.div(P_D * balances[j] * numTokens, invariant, roundUp);
-            }
+            uint256 P_D = balance0 * numTokens;
+            P_D = Math.div(P_D * balance1 * numTokens, invariant, roundUp);
```
RedFox20, line 59: I did not want to simplify & optimize math yet, to avoid potential mistakes

contracts/amm/math/StableMath.sol
```diff
@@ -51,8 +42,8 @@ library StableMath {
 
         // We support rounding up or down.
 
-        uint256 numTokens = balances.length;
-        uint256 sum = sumElements(balances);
+        uint256 numTokens = 2;
```
axic, line 45: Gas wise making this a contract/file-level constant could ensure it is always used as a constant and not a stack item.

contracts/amm/math/StableMath.sol
```diff
@@ -112,30 +100,27 @@ library StableMath {
...
         // Given that we need to have a greater final balance out, the invariant needs to be rounded up
-        uint256 invariant = _calculateInvariant(amplificationParameter, balances, true);
-
-        balances[tokenIndexIn] = balances[tokenIndexIn] + tokenAmountIn;
+        uint256 invariant = _calculateInvariant(amplificationParameter, balance0, balance1, true);
 
         uint256 finalBalanceOut = _getTokenBalanceGivenInvariantAndAllOtherBalances(
             amplificationParameter,
-            balances,
+            tokenIndexIn == 0 ? balance0 + tokenAmountIn : balance0,
```
axic, line 107: Why not just have `balance0 += tokenAmountIn`, etc outside?

RedFox20: It's not a good habit to change input parameter values. Can lead to really annoying bugs in my experience, especially when working in a large team

test/amm/StableMath.test.ts
```diff
@@ -87,7 +87,7 @@ describe('StableMath', () =>
 
   it('invariant reverts if it does not converge', async () =>
   {
-    await expect(mockMath.invariant(amp(5000), [fp(0.00001), fp(1200000), fp(300)], true))
+    await expect(mockMath.invariant(amp(50000), [fp(1.777777777777), fp(18181818181818181818)], true))
```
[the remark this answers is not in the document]
RedFox20: To make it really hard to converge. The difference between those numbers is really huge, and the large amp value makes it worse. So the loop will iterate 255 times an then revert with no convergence

contracts/amm/math/StableMath.sol
```diff
@@ -12,21 +12,11 @@ library StableMath {
     uint256 internal constant _MAX_AMP = 5000;
     uint256 internal constant _AMP_PRECISION = 1e3;
 
-    uint256 internal constant _MAX_STABLE_TOKENS = 5;
-
```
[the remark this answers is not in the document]
RedFox20: I'll add that in quick 💯

mijovic (approved):

contracts/amm/math/StableMath.sol
axic: Yeah, except gas wise it makes sense 😬

contracts/amm/math/StableMath.sol
```diff
@@ -112,30 +101,27 @@ library StableMath {
...
-        uint256 invariant = _calculateInvariant(amplificationParameter, balances, true);
-
-        balances[tokenIndexIn] = balances[tokenIndexIn] + tokenAmountIn;
+        uint256 invariant = _calculateInvariant(amplificationParameter, balance0, balance1, true);
 
         uint256 finalBalanceOut = _getTokenBalanceGivenInvariantAndAllOtherBalances(
             amplificationParameter,
-            balances,
+            tokenIndexIn == 0 ? balance0 + tokenAmountIn : balance0,
+            tokenIndexIn == 0 ? balance1 : balance1 + tokenAmountIn,
```
axic, line 109: Realised that probably making this one `tokenIndexIn == 1 ? balance1 + ...` would have improved readability for me, but it is arguable.

RedFox20: I think I'm going to try and do that soon. We only need one parameter for tokenIndexIn, since outIndex is implied. So this will make a lot of similar code much nicer 👍🏻

PR #556, opened 2022-03-15
fix: Update NatSpec comments

StefanVitor:


contracts/ITempusPool.sol
```diff
@@ -204,6 +206,7 @@ interface ITempusPool is ITempusFees, IVersioned {
     ) external view returns (uint256);
 
     /// Gets the amount of Principals and Yields for exact amountOut(YBT/BT)
+    ///       NOTE This function can be called only before maturity
```
mijovic, line 209:
```suggestion
    /// This function can be called only before maturity
```

contracts/ITempusPool.sol
```diff
@@ -204,6 +206,7 @@ interface ITempusPool is ITempusFees, IVersioned {
     ) external view returns (uint256);
 
     /// Gets the amount of Principals and Yields for exact amountOut(YBT/BT)
```
mijovic, line 208:
```suggestion
    /// @dev Gets the amount of Principals and Yields for exact amountOut(YBT/BT)
```

mijovic:
Please rebase and squash commits. We need to get rid of merge commit from master, as we never merge mater to branches, we always rebase

mijovic (approved):

PR #557, opened 2022-03-17
Amm stablemath tests

RedFox20:


test/amm/StableMath.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+    for (let i = 0; i < actualArr.length; ++i) {
+      expectEqual(actualArr[i], expectArr[i]);
+    }
+  }
+}
+
+describe.only('StableMath', () =>
+{
+  let mock:Contract;
```
mijovic, line 50: Maybe?
```suggestion
  let mockMath:Contract;
```

test/amm/StableMath.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  it('tokenOutGivenExactBptIn', async () =>
+  {
+    const equal = (...args:any[]) => expectEquals(mock.tokenOutGivenExactBptIn, calcTokenOutGivenExactBptIn, ...args);
+    await equal(amp(100), /*balances*/[fp(10), fp(11)], /*tokenIndex*/0, fp(100), fp(10000), fp(0.1));
+  });
+
+  it('tokensOutGivenExactBptIn', async () =>
+  {
+    const equal = (...args:any[]) => expectEquals(mock.tokensOutGivenExactBptIn, calcTokensOutGivenExactBptIn, ...args);
```
mijovic, line 112: How do we make sure that implementation in typescript `StableMath.ts` is correct?
Probably we need to add some actual tests to it

test/amm/StableMath.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  it('inGivenOut', async () =>
+  {
+    const equal = (...args:any[]) => expectEquals(mock.inGivenOut, calcInGivenOut, ...args);
+    await equal(amp(100), /*balances*/[fp(10), fp(12)], /*indexIn*/0, /*indexOut*/1, /*tokenAmountOut*/fp(1));
+  });
+
+  it('outGivenIn', async () =>
+  {
+    const equal = (...args:any[]) => expectEquals(mock.outGivenIn, calcOutGivenIn, ...args);
+    await equal(amp(10), /*balances*/[fp(10), fp(11)], /*indexIn*/0, /*indexOut*/1, /*tokenAmountOut*/fp(1));
```
mijovic, line 83: I think this is token in
```suggestion
    await equal(amp(10), /*balances*/[fp(10), fp(11)], /*indexIn*/0, /*indexOut*/1, /*tokenAmountIn*/fp(1));
```

mijovic (commented):
I have a couple of more tests to suggest.
- balances are not close enough, and then we want them to diverge even more. 
```
await equal(amp(10), /*balances*/[fp(10), fp(100)], /*indexIn*/1, /*indexOut*/0, /*tokenAmountIn*/fp(50));
```

- same case, with adding more "balanced" swap (by adding token which has less balance)
```
await equal(amp(10), /*balances*/[fp(10), fp(100)], /*indexIn*/0, /*indexOut*/1, /*tokenAmountIn*/fp(5));
```

I would test both of these on amp being 1, 5, 20, 50, and 90.

test/amm/StableMath.test.ts
RedFox20: Thanks, fixed 👍🏻

test/amm/StableMath.test.ts
RedFox20: Applied 💯

RedFox20:
> I have a couple of more tests to suggest.
> I would test both of these on amp being 1, 5, 20, 50, and 90.

I added new test cases for `inGivenOut` and `outGivenIn`

mijovic:
This looks good to me. Can you just squash some of the commits and we can merge?

mijovic (approved):

PR #558, opened 2022-03-17
ci: Add Echidna assertion test mode

StefanVitor:


.circleci/config.yml
```diff
@@ -184,33 +184,58 @@ jobs:
...
+          pkg-manager: yarn
+      - run:
+          name: "Run Echidna assertion lido pool"
+          no_output_timeout: 30m
+          command: |
+            yarn echidna-test-assertion:pool:lido
+
+  echidna-assertion-yearn-pool:
+    docker:
+      - image: stefantempus/tempus-echidna:latest
```
mijovic, line 228: Would be nice to have tempus/echidna

mijovic (commented):
Looks good now with docker optimization

.circleci/config.yml
StefanVitor: I wanted that and initially, but I misread that only paid users could create "organization" on docker hub. However, it's possible that and the free users create an organization with three members, so I now created 'tempusfinance' (only 'tempus` is occupied). We will see how this with docker hub works in next period, and maybe we will have to upgrade on payed version (it's about 300$ per year for 5 user).

.circleci/config.yml
```diff
@@ -184,33 +184,58 @@ jobs:
...
-          name: "Run Echidna"
+          name: "Run Echidna yearn pool"
           no_output_timeout: 30m
           command: |
-            yarn echidna-test:pool:lido
             yarn echidna-test:pool:yearn
 
+  echidna-assertion-lido-pool:
+    docker:
+      - image: stefantempus/tempus-echidna:latest
```
mijovic, line 215: Here it is still the same

.circleci/config.yml
```diff
@@ -184,33 +184,58 @@ jobs:
...
+          pkg-manager: yarn
+      - run:
+          name: "Run Echidna assertion lido pool"
+          no_output_timeout: 30m
+          command: |
+            yarn echidna-test-assertion:pool:lido
+
+  echidna-assertion-yearn-pool:
+    docker:
+      - image: stefantempus/tempus-echidna:latest
```
mijovic, line 228: And here as well

StefanVitor: Oops, forgot those two, I also rebased after the previous merge, so I think that now everything is fine.

mijovic (approved):

PR #559, opened 2022-03-19
ci: Add docker for MythX jobs

StefanVitor:


.circleci/config.yml
```diff
@@ -122,63 +122,36 @@ jobs:
...
           name: "Run mythx"
           no_output_timeout: 30m
           command: yarn mythx-amm
 
   runMythX-Misc:
-    executor:
-      name: node/default
-      tag: '15.11'
+    docker:
+      - image: stefantempus/tempus-mythx:latest
```
mijovic, line 150: Can we somehow name it tempus/mythx:latest?

mijovic:
Pretty nice optimization

.circleci/config.yml
StefanVitor: I wanted that and initially, but I misread that only paid users could create "organization" on docker hub. However, it's possible that and the free users create an organization with three members, so I now created 'tempusfinance' (only 'tempus` is occupied). We will see how this with docker hub works in next period, and maybe we will have to upgrade on payed version (it's about 300$ per year for 5 user).

mijovic (approved):

PR #560, opened 2022-03-20
[PoolTestFixture] support more than 3 signers

yuval-weiss:


test/pool-utils/AaveTestPool.ts
```diff
@@ -20,12 +20,11 @@ export class AaveTestPool extends PoolTestFixture {
...
     await this.aave.contract.setFailNextDepositOrRedeem(true);
   }
-  async getSigners(): Promise<[Signer,Signer,Signer]> {
+  async getSigners(): Promise<Signer[]> {
     if (this.integration) {
       // TODO: implement for integration tests
     } else {
-      const [owner,user,user2] = await ethers.getSigners();
-      return [owner,user,user2];
+      return await ethers.getSigners();
```
mijovic, line 27: Isn't await here not needed? We return promise anyway.

mijovic (commented):
Can we have only one function instead of multiple copies everywhere?

test/pool-utils/AaveTestPool.ts
yuval-weiss: It's not needed but imo it improves readability to include the await anyway (because that way it's clear that `ethers.getSigners` returns a promise)

yuval-weiss:
> Can we have only one function instead of multiple copies everywhere?

yup, done.

test/pool-utils/LidoTestPool.ts
```diff
@@ -23,8 +23,9 @@ export class LidoTestPool extends PoolTestFixture {
   async forceFailNextDepositOrRedeem(): Promise<void> {
     await this.lido.contract.setFailNextDepositOrRedeem(true);
   }
-  async getSigners(): Promise<[Signer,Signer,Signer]> {
+  async getSigners(): Promise<Signer[]> {
```
mijovic, line 26: I know it is annoying, but would be nice to move it to some kind of utility file if possible.
Otherwise everything looks good to me.

test/pool-utils/PoolTestFixture.ts
```diff
@@ -196,14 +196,16 @@ export abstract class PoolTestFixture {
...
+   * Deposit BackingTokens into the UNDERLYING pool and receive YBT
    */
-  abstract getSigners(): Promise<[Signer,Signer,Signer]>;
+  abstract deposit(user:Signer, amount:number): Promise<void>;
 
   /**
-   * Deposit BackingTokens into the UNDERLYING pool and receive YBT
+   * Gets the owner and users for testing
    */
-  abstract deposit(user:Signer, amount:number): Promise<void>;
```
StefanVitor, line 206: Reorder functions `getSigners` and `deposit` because for function `deposit` nothing is change.

yuval-weiss: but `getSigners` is not an abstract method anymore (I added an implementation for it), so I moved it down...

StefanVitor: Okay, my mistake, I saw that `deposit` is not changed, I didn't figure about `getSigners`

StefanVitor (approved):

RedFox20 (approved):
Looks really good, great improvement 💯

PR #561, opened 2022-03-21
ci: Check whether the YBT amount after redeem will be less or equal to YBT amount that is deposited

StefanVitor:


contracts/echidna/EchidnaTempusPool.sol
```diff
@@ -100,6 +109,32 @@ abstract contract EchidnaTempusPool {
...
+    ) internal {
+        uint256 currentPrincipalAmount = IERC20(address(tempusPool.principalShare())).balanceOf(
+            convertAddressToLimitedSet(fromRecipient)
+        );
+        uint256 currentYieldAmount = IERC20(address(tempusPool.yieldShare())).balanceOf(
+            convertAddressToLimitedSet(fromRecipient)
+        );
+
+        if (currentPrincipalAmount == 0 && currentYieldAmount == 0) {
+            _depositYieldBearing(yieldTokenAmount, fromRecipient);
```
mijovic, line 125: this should return amount of minted shares, which can be used later for withdrawal

mijovic: Instead of having sharesAmount as parameter

StefanVitor: But if you think that minted amount is enough for the check, I'll change it to that.

mijovic: Also, this will eliminate the condition for principal and yield amount to be 0

mijovic (approved):

PR #563, opened 2022-03-23
deployments: Validate deployment scripts

StefanVitor:


scripts/deploy.pool.config.ts
```diff
@@ -15,6 +15,7 @@ interface YBTConfig {
 }
 
 interface TokenNameSymbol {
+  address:string,
```
axic, line 18: Good addition.

scripts/deploy.pool.config.ts
```diff
@@ -190,21 +194,23 @@ async function deployAmm(config:Config, deployerPrivateKey:string): Promise<Conf
...
+  const maturityTimestamp = Date.parse(config.pool.maturity) / 1000;
+
   const ammConstructorArgs = [
     config.amm.vault,
     config.amm.lp.name,
     config.amm.lp.symbol,
-    config.pool.address,
-    config.amm.initialAmplificationFactor,
-    config.amm.finalAmplificationFactor,
+    [config.principal.address, config.yield.address],
```
axic, line 203: Hmm, so I guess the old Lido script was never updated?

scripts/deploy.pool.config.ts
```diff
@@ -190,21 +194,23 @@ async function deployAmm(config:Config, deployerPrivateKey:string): Promise<Conf
...
+    config.amm.initialAmplificationFactor * AMP_PRECISION,
+    config.amm.finalAmplificationFactor * AMP_PRECISION,
+    maturityTimestamp,
     parseDecimal(config.amm.swapFee, 18),
```
axic, line 207: And here instead of 18 could use `AMP_PRECISION`.

scripts/deploy.pool.config.ts
```diff
@@ -190,21 +194,23 @@ async function deployAmm(config:Config, deployerPrivateKey:string): Promise<Conf
...
   const ammConstructorArgs = [
     config.amm.vault,
     config.amm.lp.name,
     config.amm.lp.symbol,
-    config.pool.address,
-    config.amm.initialAmplificationFactor,
-    config.amm.finalAmplificationFactor,
+    [config.principal.address, config.yield.address],
+    config.amm.initialAmplificationFactor * AMP_PRECISION,
+    config.amm.finalAmplificationFactor * AMP_PRECISION,
```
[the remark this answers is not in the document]
StefanVitor: No, it was changed and added on this commit - https://github.com/tempus-finance/tempus-protocol/commit/22523276b1b374001982ff4a087a6d55fe72907e . From that moment, amplification start and amplification end values in setters (AMM constructor and `startAmplificationParameterUpdate` function) are always sent with `* 1000`.

scripts/deploy.pool.config.ts
StefanVitor: Thanks

scripts/deploy.pool.config.ts
StefanVitor: I'm not sure that I understand this question, could you explain me more detail what you mean by this?

scripts/deploy.pool.config.ts
StefanVitor: Explained in the previous comment.

scripts/deploy.pool.config.ts
```diff
@@ -15,6 +15,7 @@ interface YBTConfig {
 }
 
 interface TokenNameSymbol {
```
yuval-weiss, line 17:
```suggestion
interface TokenInfo {
```

StefanVitor: Agree with this.

scripts/deploy.pool.config.ts
yuval-weiss: @axic `swapFee` is always 18 decimals

scripts/deploy.pool.config.ts
axic: Ah never mind, this was changed as part of the AMM refactoring, but the deployment script was not updated at the time.

yuval-weiss (approved):

PR #576, opened 2022-03-29
amm: v2

mijovic:
TODO:
- [x] deadline in Swaps
- [x] NatSpec
- [x] Test infra update
- [x] Add recipient as param in AMM
- [x] Implement not implemented methods
- [x] Rebase to latest StableMath interface
- [x] Update deployment scripts
- [x] Fix MythX job for AMM


contracts/ITempusController.sol
```diff
@@ -68,14 +68,6 @@ interface ITempusController {
         bool isBackingToken
     ) external payable;
 
-    /// @dev Adds liquidity to tempusAMM with ratio of shares that is equal to ratio in AMM
-    /// @param tempusAMM Tempus AMM to provide liquidity to
-    /// @param sharesAmount Amount of shares to be used to provide liquidity, one of the sahres will be partially used
-    /// @notice If sharesAmount is 100 and amm balances ratio is 1 principal : 10 yields 90 principal will be "unused"
-    ///         So, liquidity will be provided with 10 principals and 100 yields
-    /// @notice msg.sender needs to approve Controller for both Principals and Yields for @param sharesAmount
-    function provideLiquidity(ITempusAMM tempusAMM, uint256 sharesAmount) external;
```
mijovic, line 77: Useless, one can just LP to the amm instead

contracts/stats/Stats.sol
```diff
@@ -118,13 +118,17 @@ contract Stats is ChainlinkTokenPairPriceFeed, Versioned {
     {
         uint256 shares = estimatedMintedShares(tempusPool, amount, isBackingToken);
 
-        (IERC20[] memory ammTokens, uint256[] memory ammBalances, ) = tempusAMM.getVault().getPoolTokens(
-            tempusAMM.getPoolId()
+        uint256 ammBalance0 = tempusAMM.token0().balanceOf(address(tempusAMM));
+        uint256 ammBalance1 = tempusAMM.token1().balanceOf(address(tempusAMM));
+        
+        (uint256 ammLPAmount0, uint256 ammLPAmount1) = AMMBalancesHelper.getLiquidityProvisionSharesAmounts(
```
RedFox20, line 124: Since we use `LP` everywhere. Maybe we should rename this to `AMMUtils.getLPSharesAmounts()` and spell it out in the documentation instead?

test/TempusController.test.ts
```diff
@@ -32,8 +32,8 @@ describeForEachPool("TempusController", (testPool:PoolTestFixture) =>
 
   async function getAMMBalancesRatio(): Promise<BigNumber>
   {
-    const principals = await pool.principalShare.balanceOf(amm.vault.address);
-    const yields = await pool.yieldShare.balanceOf(amm.vault.address);
+    const principals = await pool.principalShare.balanceOf(amm.address);
```
RedFox20, line 35: So I see we are not storing them in vault anymore. What about the new Vault PR?

test/amm/TempusAMM.test.ts
```diff
@@ -79,9 +79,11 @@ describeForEachPool("TempusAMM", (testFixture:PoolTestFixture) =>
     // mine a block in case the current test case has automining set to false (otherwise expect functions would fail...)
     await evmMine();
 
+    console.log(tokenIn.fromBigNum(preSwapTokenInBalance), tokenIn.fromBigNum(preSwapTokenOutBalance));
     const postSwapTokenInBalance:BigNumber = await tokenIn.contract.balanceOf(owner.address);
```
RedFox20, line 83: Why are we doing `tokenIn.contract.balanceOf` instead of `tokenIn.balanceOf` which returns an already parsed NumberOrString ?

test/amm/TempusAMM.test.ts
```diff
@@ -355,7 +333,7 @@ describeForEachPool("TempusAMM", (testFixture:PoolTestFixture) =>
     expect(+await tempusAMM.getRate()).to.be.within(1.005975, 1.0060);
   });
 
-  it("test swaps principal in with balances aligned with Interest Rate", async () =>
+  it.only("test swaps principal in with balances aligned with Interest Rate", async () =>
```
RedFox20, line 336: Need to remove `.only` at some point

test/amm/TempusAMM.test.ts
```diff
@@ -260,8 +240,6 @@ describeForEachPool("TempusAMM", (testFixture:PoolTestFixture) =>
     // stop update
     await tempusAMM.stopAmplificationUpdate();
     await tempusAMM.provideLiquidity(owner, 100, 1000, TempusAMMJoinKind.INIT);
-    const ampInv = await tempusAMM.getLastInvariant();
-    expect(ampInv.amplification).to.equal(35000);
```
RedFox20, line 264: We need an alternative check here maybe?

test/amm/TempusAMM.test.ts
```diff
@@ -216,18 +206,8 @@ describeForEachPool("TempusAMM", (testFixture:PoolTestFixture) =>
...
-    expect(amplificationParams.value).to.be.equal(ampInv.amplification);
     expect(amplificationParams.isUpdating).to.be.true;
-    expect(ampInv.invariant).to.be.within(200 / 1.11, 200 / 1.09);
-    
-    // move half period of pool duration
-    await testFixture.setTimeRelativeToPoolStart(0.5);
-    await testFixture.setInterestRate(1.05);
-    await testFixture.amm.provideLiquidity(owner, 100, 1000, 1);
-    ampInv = await testFixture.amm.getLastInvariant();
-    expect(ampInv.invariant).to.be.within(400 / (1.1 / 1.049), 400 / (1.1 / 1.051));
```
RedFox20, line 230: Why was all of this removed? Because we no longer have `getLastInvariant()` ?

test/utils/TempusAMM.ts
```diff
@@ -68,37 +55,22 @@ export class TempusAMM extends ContractBase {
     amplificationEndTime: number,
     swapFeePercentage: number,
   ): Promise<TempusAMM> {
-
-    const mockedWETH = await TempusAMM.createMock();
     
-    const authorizer = await ContractBase.deployContract("@balancer-labs/v2-vault/contracts/Authorizer.sol:Authorizer", owner.address);
-    const vault = await ContractBase.deployContract("@balancer-labs/v2-vault/contracts/Vault.sol:Vault", authorizer.address, mockedWETH.address, 3 * MONTH, MONTH);
```
RedFox20, line 75: Nice. Removing this mess is great

test/utils/TempusAMM.ts
```diff
@@ -146,27 +118,14 @@ export class TempusAMM extends ContractBase {
...
-      assets: assets.map(({ address }) => address),
-      maxAmountsIn: initialBalances,
-      userData: initUserData,
-      fromInternalBalance: false
-    };
+    await this.token0.approve(from, this.address, token0Balance);
+    await this.token1.approve(from, this.address, token1Balance);
   
-    await this.vault.connect(from).joinPool(poolId, from.address, from.address, joinPoolRequest);
+    if (joinKind == TempusAMMJoinKind.INIT) {
```
RedFox20, line 124: This looks so much simpler!

contracts/amm/interfaces/ITempusAMM.sol
```diff
@@ -1,34 +1,33 @@
...
-
-    function transferFrom(
-        address sender,
-        address recipient,
-        uint256 amount
-    ) external returns (bool);
+    function swap(
+        IPoolShare tokenIn, 
+        uint256 amount, 
+        uint256 slippageParam,
```
RedFox20, line 27: I think we need to document these functions, what is `slippageParam` ?

contracts/TempusController.sol
```diff
@@ -73,12 +62,12 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
         requireRegistered(address(tempusAMM));
         requireRegistered(address(tempusPool));
 
-        IERC20 principalShares = IERC20(address(tempusPool.principalShare()));
-        IERC20 yieldShares = IERC20(address(tempusPool.yieldShare()));
+        IPoolShare principalShares = tempusPool.principalShare();
```
RedFox20, line 65: This is great 🤤

contracts/TempusController.sol
```diff
@@ -73,12 +62,12 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
...
+        IPoolShare yieldShares = tempusPool.yieldShare();
 
         uint256 swapAmount = _deposit(tempusPool, tokenAmount, isBackingToken);
         uint256 minReturn = swapAmount.mulfV(minTYSRate, tempusPool.backingTokenONE());
-        swap(tempusAMM, swapAmount, yieldShares, principalShares, minReturn, deadline);
+        swap(tempusAMM, swapAmount, yieldShares, minReturn, deadline);
 
         // At this point all TYS must be swapped for TPS
         uint256 principalsBalance = principalShares.balanceOf(address(this));
```
RedFox20, line 73: Also, I think `safeTransfer()` can be replaced with regular transfer, because we trust the token

contracts/TempusController.sol
```diff
@@ -251,64 +238,28 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
...
-        });
-
-        IVault.FundManagement memory fundManagement = IVault.FundManagement({
-            sender: address(this),
-            fromInternalBalance: false,
-            recipient: payable(address(this)),
-            toInternalBalance: false
-        });
-        vault.swap(singleSwap, fundManagement, minReturn, deadline);
+        tokenIn.safeIncreaseAllowance(address(tempusAMM), swapAmount);
```
RedFox20, line 246: We trust the token, so I believe we can just use increaseAllowance instead

contracts/TempusController.sol
```diff
@@ -251,64 +238,28 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
...
-        });
-
-        IVault.FundManagement memory fundManagement = IVault.FundManagement({
-            sender: address(this),
-            fromInternalBalance: false,
-            recipient: payable(address(this)),
-            toInternalBalance: false
-        });
-        vault.swap(singleSwap, fundManagement, maxSpendAmount, deadline);
+        tokenIn.safeIncreaseAllowance(address(tempusAMM), maxSpendAmount);
```
RedFox20, line 260: Here as well, this is a trusted token, so we can just use increaseAllowance
All we need to do is ensure that `tokenIn` is validated in TempusAMM

contracts/TempusController.sol
```diff
@@ -317,64 +268,48 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
...
 
         uint256 mintedShares = _deposit(tempusPool, tokenAmount, isBackingToken);
 
-        uint256[] memory sharesUsed = _provideLiquidity(
-            address(this),
-            vault,
-            poolId,
-            ammTokens,
-            ammBalances,
+        (uint256 sharesUsed0, uint256 sharesUsed1) = _provideLiquidity(
```
RedFox20, line 275: I think `sharesUsed0` should have a better name `principalsUsed` ?

contracts/TempusController.sol
```diff
@@ -317,64 +268,48 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
...
+            sharesAmount
+        );
 
-        IVault.JoinPoolRequest memory request = IVault.JoinPoolRequest({
-            assets: ammTokens,
-            maxAmountsIn: ammLiquidityProvisionAmounts,
-            userData: abi.encode(uint8(ITempusAMM.JoinKind.EXACT_TOKENS_IN_FOR_BPT_OUT), ammLiquidityProvisionAmounts),
-            fromInternalBalance: false
-        });
+        tempusAMM.token0().safeIncreaseAllowance(address(tempusAMM), ammLPAmount0);
```
RedFox20, line 308: Here as well, is `safeIncreaseAllowance` needed?

contracts/TempusController.sol
```diff
@@ -537,49 +472,18 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
...
         uint256 principalAmountOutMin,
         uint256 yieldAmountOutMin,
-        address recipient,
-        bool toInternalBalances
+        address recipient
     ) private {
         require(tempusAMM.transferFrom(msg.sender, address(this), lpTokensAmount), "LP token transfer failed");
 
-        uint256[] memory minAmountsOut = getAMMOrderedAmounts(
+        (uint256 minAmountOut0, uint256 minAmountOut1) = getAMMOrderedAmounts(
```
RedFox20, line 479: `getAMMOrderedAmounts()` name doesn't make much sense because we always have token0 == principals now

contracts/TempusController.sol
```diff
@@ -677,33 +585,23 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
         private
         view
         returns (
-            IVault vault,
-            bytes32 poolId,
-            IERC20[] memory ammTokens,
-            uint256[] memory ammBalances
+            uint256 ammBalance0,
```
RedFox20, line 588: Instead of `ammBalance0` we could probably make use of principals/yields naming

contracts/TempusController.sol
```diff
@@ -677,33 +585,23 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
...
         ITempusPool tempusPool,
         uint256 principalAmount,
         uint256 yieldAmount
-    ) private view returns (uint256[] memory) {
-        IVault vault = tempusAMM.getVault();
-        (IERC20[] memory ammTokens, , ) = vault.getPoolTokens(tempusAMM.getPoolId());
-        uint256[] memory amounts = new uint256[](2);
-        (amounts[0], amounts[1]) = (address(tempusPool.principalShare()) == address(ammTokens[0]))
+    ) private view returns (uint256, uint256) {
+        return (tempusPool.principalShare() == tempusAMM.token0())
```
RedFox20, line 603: This function does nothing. It is always `(principalAmount, yieldAmount)`

contracts/amm/TempusAMM.sol
```diff
@@ -70,17 +67,9 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
     // not change throughout its lifetime, and store the corresponding scaling factor for each at construction time.
     // These factors are always greater than or equal to one: tokens with more than 18 decimals are not supported.
 
-    uint256 internal immutable _scalingFactor0;
-    uint256 internal immutable _scalingFactor1;
-
-    // To track how many tokens are owed to the Vault as protocol fees, we measure and store the value of the invariant
-    // after every join and exit. All invariant growth that happens between join and exit events is due to swap fees.
-    uint256 internal _lastInvariant;
+    uint256 internal immutable _scalingFactor;
```
RedFox20, line 70: We should probably remove `_` prefix from internal variables. Because we never use them on things like `Math.div()` etc

contracts/amm/TempusAMM.sol
```diff
@@ -92,96 +81,77 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-        assert(_TOTAL_TOKENS == 2);
+        require(amplificationStartValue >= _MIN_AMPLIFICATION, "min amp");
+        require(amplificationStartValue <= _MAX_AMPLIFICATION, "max amp");
 
-        _require(amplificationStartValue >= _MIN_AMPLIFICATION, Errors.MIN_AMP);
-        _require(amplificationStartValue <= _MAX_AMPLIFICATION, Errors.MAX_AMP);
+        swapFeePercentage = swapFeePerc;
 
-        (IPoolShare t0, IPoolShare t1) = (tokens[0], tokens[1]);
-        require(t0 < t1, "token0 < token1 address");
```
RedFox20, line 129: Why did we remove this require here? It was crucial before?

contracts/amm/TempusAMM.sol
```diff
@@ -92,96 +81,77 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-
-        amount = _subtractSwapFeeAmount(amount);
-        balances.mul(_getTokenRatesStored(), _TEMPUS_SHARE_PRECISION);
-        uint256 rateAdjustedSwapAmount = (amount * tokenIn.getPricePerFullShareStored()) / _TEMPUS_SHARE_PRECISION;
-
-        uint256 amountOut = StableMath._calcOutGivenIn(currentAmp, balances, indexIn, indexOut, rateAdjustedSwapAmount);
-        amountOut = (amountOut * _TEMPUS_SHARE_PRECISION) / tokenOut.getPricePerFullShareStored();
-
-        return amountOut;
+        (IPoolShare tokenOut, bool firstIn) = (tokenIn == token0) ? (token1, true) : (token0, false);
```
RedFox20, line 120: Instead of making `firstIn` a variable, could it be cheaper to just call `outGivenIn` with firstIn:  `tokenIn == token0` expression? Just a question, not exactly a suggestion

contracts/amm/TempusAMM.sol
```diff
@@ -226,480 +196,269 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-        balances.mul(tokenRates, _TEMPUS_SHARE_PRECISION);
-
-        uint256 protocolSwapFeePercentage = getSwapFeePercentage();
-        if (_isNotPaused()) {
-            // Update current balances by subtracting the protocol fee amounts
-            balances.sub(_getDueProtocolFeeAmounts(balances, protocolSwapFeePercentage));
-        }
+        (uint256 balance0, uint256 balance1) = _getRateAdjustedBalancesStored();
+        
+        token0Out = token0Out.mulDown(_scalingFactor).mulfV(
```
RedFox20, line 201: Do we need `scalingFactor` now?

contracts/amm/TempusAMM.sol
```diff
@@ -226,480 +196,269 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
         returns (uint256 token0Out, uint256 token1Out)
     {
         // We don't need to scale balances down here
         // as calculation for amounts out is based on btpAmountIn / totalSupply() ratio
         // Adjusting balances with rate, and then undoing it would just cause additional calculations
-        (, uint256[] memory balances, ) = getVault().getPoolTokens(getPoolId());
-        uint256[] memory amountsOut = StableMath._calcTokensOutGivenExactBptIn(balances, bptAmountIn, totalSupply());
-        return (amountsOut[0], amountsOut[1]);
+        return StableMath.tokensOutFromBptIn(
+            token0.balanceOf(address(this)),
```
RedFox20, line 229: We use `token0.balanceOf(address(tempusAMM))` a lot. Could we maybe create some view functions with descriptive naming to reduce repetition? (assuming it would get inlined)

contracts/amm/TempusAMM.sol
```diff
@@ -226,480 +196,269 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-        uint256[] memory tokenRates = _getTokenRatesStored();
-        balances.mul(tokenRates, _TEMPUS_SHARE_PRECISION);
-        amountsIn.mul(tokenRates, _TEMPUS_SHARE_PRECISION);
+    function getExpectedLPTokensForTokensIn(
+        uint256 token0AmountIn, 
+        uint256 token1AmountIn
+    ) external view override returns (uint256) {
+        (uint256 balance0, uint256 balance1) = _getRateAdjustedBalancesStored();
+        
+        token0AmountIn = token0AmountIn.mulDown(_scalingFactor).mulfV(
```
RedFox20, line 242: This is a duplicate of `getSwapAmountToEndWithEqualShares`, so maybe we could use a function for it.

Also, Tempus shares are always 1e18, so we could make use of this knowledge and reduce some gas usage in these operations later

contracts/amm/TempusAMM.sol
```diff
@@ -226,480 +196,269 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-    function _onSwapGivenIn(
-        SwapRequest memory swapRequest,
-        uint256 balanceTokenIn,
-        uint256 balanceTokenOut
-    ) internal virtual override whenNotPaused returns (uint256) {
-        (uint256[] memory balances, uint256 indexIn, uint256 indexOut) = _getSwapBalanceArrays(
-            swapRequest,
-            balanceTokenIn,
-            balanceTokenOut
+        uint256 amountIn0 = amountToken0.mulDown(_scalingFactor).mulfV(
```
RedFox20, line 268: This is the third duplication of this same expression pair

contracts/amm/TempusAMM.sol
```diff
@@ -226,480 +196,269 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-    ) private view returns (uint256, uint256[] memory) {
-        (uint256[] memory amountsIn, uint256 minBPTAmountOut) = userData.exactTokensInForBptOut();
-        InputHelpers.ensureInputLengthMatch(_TOTAL_TOKENS, amountsIn.length);
-
-        _upscaleArray(amountsIn, scalingFactors);
-        amountsIn.mul(tokenRates, _TEMPUS_SHARE_PRECISION);
+    function join(uint256 amountToken0, uint256 amountToken1, uint256 minLpTokensOut, address recipient) external override whenNotPaused {
+        (uint256 balance0, uint256 balance1) = _getRateAdjustedBalances();
+        
+        uint256 amountIn0 = amountToken0.mulDown(_scalingFactor).mulfV(
```
RedFox20, line 289: Fourth instance where this is duplicated

contracts/amm/TempusAMM.sol
```diff
@@ -226,480 +196,269 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
+            balance1,
+            amountIn0,
+            amountIn1,
             totalSupply(),
-            getSwapFeePercentage()
+            swapFeePercentage
         );
 
-        _require(bptAmountOut >= minBPTAmountOut, Errors.BPT_OUT_MIN_AMOUNT);
+        require(lpTokensOut >= minLpTokensOut, "slippage");
```
RedFox20, line 308: Maybe a bit more description here is needed, especially for frontend guys

contracts/amm/TempusAMM.sol
```diff
@@ -226,480 +196,269 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
+    function exitGivenLpIn(uint256 lpTokensIn, uint256 minAmountOut0, uint256 minAmountOut1, address recipient) external override {
         // This exit function is the only one that is not disabled if the contract is paused: it remains unrestricted
         // in an attempt to provide users with a mechanism to retrieve their tokens in case of an emergency.
         // This particular exit function is the only one that remains available because it is the simplest one, and
         // therefore the one with the lowest likelihood of errors.
 
-        uint256 bptAmountIn = userData.exactBptInForTokensOut();
-        // Note that there is no minimum amountOut parameter: this is handled by `IVault.exitPool`.
+        (uint256 amountOut0, uint256 amountOut1) = StableMath.tokensOutFromBptIn(
+            token0.balanceOf(address(this)),
```
RedFox20, line 322: `token0.balanceOf(address(AMM))` again, we could really use a function for it

contracts/amm/TempusAMM.sol
```diff
@@ -226,480 +196,269 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
 
-        (uint256[] memory amountsOut, uint256 maxBPTAmountIn) = userData.bptInForExactTokensOut();
-        InputHelpers.ensureInputLengthMatch(amountsOut.length, _TOTAL_TOKENS);
-        _upscaleArray(amountsOut, scalingFactors);
-
-        amountsOut.mul(tokenRates, _TEMPUS_SHARE_PRECISION);
-
+        (uint256 balance0, uint256 balance1) = _getRateAdjustedBalances();
+        
+        uint256 amountOut0 = token0AmountOut.mulDown(_scalingFactor).mulfV(
```
RedFox20, line 340: Fifth duplication of this same expression pair.

contracts/amm/TempusAMM.sol
```diff
@@ -226,480 +196,269 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-            getSwapFeePercentage()
+            swapFeePercentage
         );
-        _require(bptAmountIn <= maxBPTAmountIn, Errors.BPT_IN_MAX_AMOUNT);
-
-        return (bptAmountIn, amountsOut);
-    }
-
-    // Helpers
+        require(lpTokensIn <= maxLpTokensIn, "slippage");
```
RedFox20, line 357: Needs better description and docs

contracts/amm/TempusAMM.sol
```diff
@@ -226,480 +196,269 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-            _lastInvariant,
-            chosenTokenIndex,
-            protocolSwapFeePercentage
-        );
+        (uint256 balance0, uint256 balance1) = _getRateAdjustedBalances();
+        (uint256 currentAmp, ) = _getAmplificationParameter();
 
-        return dueProtocolFeeAmounts;
+        (uint256 amountIn, uint256 amountOut) = (swapType == SwapType.GIVEN_IN)
+            ? (amount, uint256(0))
```
RedFox20, line 382: Why not just use `0` ?

contracts/amm/TempusAMM.sol
```diff
@@ -226,480 +196,269 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-        }
 
-        // Instead of paying the protocol swap fee in all tokens proportionally, we will pay it in a single one. This
-        // will reduce gas costs for single asset joins and exits, as at most only two Pool balances will change (the
-        // token joined/exited, and the token in which fees will be paid).
+    function swap(
+        IPoolShare tokenIn, 
+        uint256 amount, 
+        uint256 slippageParam, 
+        SwapType swapType,
```
RedFox20, line 369: Maybe `SwapType type` ?

contracts/TempusController.sol
```diff
@@ -251,64 +238,28 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
...
         ITempusAMM tempusAMM,
         uint256 swapAmount,
-        IERC20 tokenIn,
-        IERC20 tokenOut,
+        IPoolShare tokenIn,
         uint256 minReturn,
         uint256 deadline
     ) private {
         require(swapAmount > 0, "Invalid swap amount.");
-        tokenIn.safeIncreaseAllowance(address(tempusAMM.getVault()), swapAmount);
```
StefanVitor, line 260: This is great.

contracts/TempusController.sol
```diff
@@ -19,6 +19,7 @@ import "./utils/Versioned.sol";
 ///      Owner is automatically set to the deployer of this contract
 contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versioned {
```
StefanVitor, line 20: Line #9 (import "./amm/interfaces/IVault.sol";) is not needed anymore. Also, "./amm/interfaces/IVault.sol" should be delete.

contracts/amm/TempusAMM.sol
axic: We still didn't change the shares to be 18 decimals.

contracts/amm/TempusAMM.sol
axic: > assuming it would get inlined

It won't neccessarily be inlined, but need to try.

contracts/amm/TempusAMM.sol
```diff
@@ -858,9 +587,13 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
 
-    function _mapSharesToIERC20(IPoolShare[2] memory tokens) private pure returns (IERC20[] memory) {
-        IERC20[] memory mapped = new IERC20[](_TOTAL_TOKENS);
-        (mapped[0], mapped[1]) = (IERC20(address(tokens[0])), IERC20(address(tokens[1])));
-        return mapped;
+    /**
+     * @dev Returns a scaling factor that, when multiplied to a token amount for `token`, normalizes its balance as if
+     * it had 18 decimals.
+     */
+    function _computeScalingFactor(IERC20 token) internal view returns (uint256) {
```
axic, line 594: Shouldn't this use `IPoolShare` now?

axic: Could potentially also inline it in the constructor given it is used once.

contracts/amm/TempusAMM.sol
```diff
@@ -226,480 +196,269 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
+        balance1 = token1.balanceOf(address(this)).mulDown(_scalingFactor);
     }
 
-    /// @dev Creates 2 element array of token rates(pricePerFullShareStored)
-    /// @return rates Array of stored token rates
-    function _getTokenRatesStored() private view returns (uint256[] memory rates) {
-        rates = new uint256[](_TOTAL_TOKENS);
-        rates[0] = token0.getPricePerFullShareStored();
-        rates[1] = token1.getPricePerFullShareStored();
+    function _addSwapFeeAmount(uint256 amount) internal view returns (uint256) {
```
axic, line 443: Could change the `internal`s to `private`.

contracts/amm/TempusAMM.sol
```diff
@@ -226,480 +196,269 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
     }
 
-    /// @dev Creates 2 element array of token rates(pricePerFullshare)
-    /// @return rates Array of token rates
-    function _getTokenRates() private returns (uint256[] memory rates) {
-        rates = new uint256[](_TOTAL_TOKENS);
-        rates[0] = token0.getPricePerFullShare();
-        // We already did updateInterestRate, so we can use stored values
-        rates[1] = token1.getPricePerFullShareStored();
+    function _getUpscaledBalances() internal view returns (uint256 balance0, uint256 balance1) {
```
axic, line 438: For readability I'd move this function before the `getRate*` functions.

contracts/amm/TempusAMM.sol
```diff
@@ -226,480 +196,269 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
                     totalSupply(),
-                    getSwapFeePercentage()
+                    swapFeePercentage
                 );
     }
 
-    // Base Pool handlers
-
-    // Swap - Two Token Pool specialization (from BaseMinimalSwapInfoPool)
+    function init(uint256 amountToken0, uint256 amountToken1) external {
```
axic, line 265: What is this function? Why is this external, not gated, not pausable?

contracts/amm/TempusAMM.sol
```diff
@@ -226,480 +196,269 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-            revert("Unhandled exit kind.");
-        }
-    }
-
-    function _exitExactBPTInForTokensOut(uint256[] memory balances, bytes memory userData)
-        private
-        view
-        returns (uint256, uint256[] memory)
-    {
+    function exitGivenLpIn(uint256 lpTokensIn, uint256 minAmountOut0, uint256 minAmountOut1, address recipient) external override {
```
axic, line 315: Should all state changing functions be pausable? This one is not.

contracts/amm/TempusAMM.sol
```diff
@@ -709,48 +468,41 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
      * @dev Begins changing the amplification parameter to `endValue` over time. The value will change linearly until
      * `endTime` is reached, when it will be `endValue`.
      */
-    function startAmplificationParameterUpdate(uint256 endValue, uint256 endTime) external authenticate {
+    function startAmplificationParameterUpdate(uint256 endValue, uint256 endTime) external onlyOwner {
```
axic, line 471: Should move this to the interface too and mark this `override`.

contracts/amm/TempusAMM.sol
```diff
@@ -709,48 +468,41 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
+        require(dailyRate <= _MAX_AMP_UPDATE_DAILY_RATE, "amp rate too high");
 
         _setAmplificationData(currentValue, endValue, block.timestamp, endTime);
     }
 
     /**
      * @dev Stops the amplification parameter change process, keeping the current value.
      */
-    function stopAmplificationParameterUpdate() external authenticate {
+    function stopAmplificationParameterUpdate() external onlyOwner {
```
axic, line 499: Should move this to the interface too and mark this `override`.

contracts/amm/TempusAMM.sol
```diff
@@ -709,48 +468,41 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
     }
 
-    function _isOwnerOnlyAction(bytes32 actionId) internal view virtual override returns (bool) {
-        return
-            (actionId == getActionId(TempusAMM.startAmplificationParameterUpdate.selector)) ||
-            (actionId == getActionId(TempusAMM.stopAmplificationParameterUpdate.selector)) ||
-            super._isOwnerOnlyAction(actionId);
-    }
-
     function getAmplificationParameter()
```
axic, line 506: Should move this to the interface too and mark this `override`.

contracts/amm/TempusAMM.sol
axic: `type` is a keyword

contracts/amm/TempusAMM.sol
axic: Also shouldn't this be pausable?

contracts/amm/TempusAMM.sol
```diff
@@ -92,96 +81,77 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
+            rateAdjustedAmountOut
+        );
         amountIn = (amountIn * _TEMPUS_SHARE_PRECISION) / shareIn.getPricePerFullShareStored();
-        amountIn = _subtractSwapFeeAmount(amountIn);
-
-        return amountIn;
+        return _addSwapFeeAmount(amountIn);
     }
 
     function getSwapAmountToEndWithEqualShares(
```
axic, line 157: Should be added to the interface and market `override`.

contracts/amm/TempusAMM.sol
```diff
@@ -92,96 +81,77 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
         IPoolShare shareIn = IPoolShare(tokenIn);
-        (, uint256[] memory balances, ) = getVault().getPoolTokens(getPoolId());
+        (uint256 balance0, uint256 balance1) = _getRateAdjustedBalancesStored();
         (uint256 currentAmp, ) = _getAmplificationParameter();
         require(shareIn == token0 || shareIn == token1, "invalid tokenIn");
         IPoolShare tokenOut = (shareIn == token0) ? token1 : token0;
-        (uint256 indexIn, uint256 indexOut) = (shareIn == token0) ? (0, 1) : (1, 0);
 
-        balances.mul(_getTokenRatesStored(), _TEMPUS_SHARE_PRECISION);
         uint256 rateAdjustedAmountOut = (amountOut * tokenOut.getPricePerFullShareStored()) / _TEMPUS_SHARE_PRECISION;
```
axic, line 144: This and a few others aren't using the fixed point helpers.

contracts/amm/TempusAMM.sol
mijovic: It is hard to believe that amm will go to unstable situation before it was initialised. Whoever deploys it, will probably init it immediately

axic: Not sure I understand, but I do not see anything in this function limiting the number of times this can be called?

contracts/stats/Stats.sol
mijovic: Done

test/TempusController.test.ts
mijovic: LP Vaults is something completely different

test/amm/TempusAMM.test.ts
mijovic: I'll do it in increasing coverage of TempusAMM

test/amm/TempusAMM.test.ts
```diff
@@ -186,18 +188,6 @@ describeForEachPool("TempusAMM", (testFixture:PoolTestFixture) =>
...
   });
 
-  it("checks amplification and invariant in multiple stages", async () =>
-  {
-    await createPools({yieldEst:0.1, duration:ONE_MONTH, amplifyStart:5, amplifyEnd:5});
-    let ampInv = await testFixture.amm.getLastInvariant();
-    expect(ampInv.invariant).to.equal(0);
-    expect(ampInv.amplification).to.equal(0);
-    await testFixture.amm.provideLiquidity(owner, 100, 1000, TempusAMMJoinKind.INIT);
-    ampInv = await testFixture.amm.getLastInvariant();
```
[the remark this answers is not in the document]
mijovic: Yes

contracts/amm/TempusAMM.sol
mijovic: We do not need sorted tokens anymore. One more good thing after removing Balancer Vault

contracts/amm/TempusAMM.sol
mijovic: Since I was already checking equality here, I was thinking to not do it twice...

contracts/amm/TempusAMM.sol
mijovic: It is not uint256 by default, and then type resolver can't resolve the type of true and false side of ternary

contracts/amm/TempusAMM.sol
mijovic: I added check `totalSupply() == 0` now, it will make sure it is called only in that case

contracts/amm/TempusAMM.sol
mijovic: There is a comment at the beginning of the function why. This is the only way to withdraw liquidity in cases of emergency

contracts/amm/TempusAMM.sol
mijovic: moved

contracts/amm/TempusAMM.sol
mijovic: moved
It shouldn't be an issue running this when paused as there is no method that reads amp and is not guarded by whenNotPaused
However, might be worth adding comment

contracts/amm/TempusAMM.sol
mijovic: Done

contracts/amm/TempusAMM.sol
mijovic: Done

contracts/amm/TempusAMM.sol
mijovic: Will double-check these

test/amm/TempusAMM.test.ts
mijovic: Exactly. I will readd tests by adding more coverage to TempusAMM

contracts/TempusController.sol
mijovic: Removed

contracts/TempusController.sol
mijovic: It is removed now

contracts/TempusController.sol
mijovic: Thanks, removed

contracts/amm/TempusAMM.sol
mijovic: Inlined, it was one line anyway

contracts/amm/TempusAMM.sol
mijovic: Done

contracts/amm/TempusAMM.sol
mijovic: I did complete reorder now. Please check again and maybe suggest some better way

contracts/amm/TempusAMM.sol
mijovic: I think I fixed it now everywhere

contracts/amm/TempusAMM.sol
```diff
@@ -58,7 +55,7 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
         uint64 endTime;
     }
 
-    AmplificationData private _amplificationData;
+    AmplificationData private amplificationData;
 
     event AmpUpdateStarted(uint256 startValue, uint256 endValue, uint256 startTime, uint256 endTime);
```
axic, line 60: Would move these events to the interface too.

contracts/amm/TempusAMM.sol
```diff
@@ -42,14 +38,15 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
     // value daily. It is possible to perform multiple amplification changes in sequence to increase this value more
     // rapidly: for example, by doubling the value every day it can increase by a factor of 8 over three days (2^3).
-    uint256 private constant _MIN_UPDATE_TIME = 1 days;
-    uint256 private constant _MAX_AMP_UPDATE_DAILY_RATE = 2;
-    uint256 private immutable _TEMPUS_SHARE_PRECISION;
-    uint256 private constant _TOTAL_TOKENS = 2;
+    uint256 private constant MIN_UPDATE_TIME = 1 days;
+    uint256 private constant MAX_AMP_UPDATE_DAILY_RATE = 2;
 
     // Tempus custom MIN_AMPLIFICATION and MAX_AMPLIFICATION constant
```
axic, line 44: Is this comment needed?

contracts/amm/TempusAMM.sol
```diff
@@ -69,703 +66,448 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-
-    enum JoinKind {
-        INIT,
-        EXACT_TOKENS_IN_FOR_BPT_OUT
-    }
-    enum ExitKind {
-        EXACT_BPT_IN_FOR_TOKENS_OUT,
-        BPT_IN_FOR_EXACT_TOKENS_OUT
-    }
+    uint256 internal immutable swapFeePercentage;
```
axic, line 71: Shouldn't this be made public? (Can be done after this PR)

contracts/amm/TempusAMM.sol
```diff
@@ -69,703 +66,448 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
 
-        _require(amplificationStartValue >= _MIN_AMPLIFICATION, Errors.MIN_AMP);
-        _require(amplificationStartValue <= _MAX_AMPLIFICATION, Errors.MAX_AMP);
+        swapFeePercentage = swapFeePerc;
 
-        (IPoolShare t0, IPoolShare t1) = (tokens[0], tokens[1]);
-        require(t0 < t1, "token0 < token1 address");
         (token0, token1) = (t0, t1);
 
         require(ERC20(address(t0)).decimals() == ERC20(address(t1)).decimals(), "token0 != token1 decimals");
```
axic, line 90: `ERC20` wrapping not needed in these places anymore.

contracts/amm/TempusAMM.sol
```diff
@@ -69,703 +66,448 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
+                firstIn,
+                rateAdjustedAmount
+            );
+
+            amountOut = scaledAmountOut.divfV(tokenOut.getPricePerFullShare(), TEMPUS_SHARE_PRECISION).divDown(
+                scalingFactor
+            );
+
+            require(amountOut >= slippageParam, "slippage");
+        } else {
```
axic, line 270: I'd add `assert(swapType == SwapType.GIVEN_OUT);` here.

contracts/amm/TempusAMM.sol
```diff
@@ -69,703 +66,448 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
     }
 
     // Amplification
 
-    /**
-     * @dev Begins changing the amplification parameter to `endValue` over time. The value will change linearly until
-     * `endTime` is reached, when it will be `endValue`.
-     */
-    function startAmplificationParameterUpdate(uint256 endValue, uint256 endTime) external authenticate {
+    function startAmplificationParameterUpdate(uint256 endValue, uint256 endTime) external override onlyOwner {
```
axic, line 463: Why do we need this helper? Can't we just make the underscore prefixed version `public`?

contracts/amm/TempusAMM.sol
mijovic: I didn't want to deal with swap fees(we might implement some different mechanics like not flat fee)

contracts/amm/TempusAMM.sol
mijovic: Agree, will add

contracts/amm/TempusAMM.sol
mijovic: Agreed

contracts/amm/TempusAMM.sol
```diff
@@ -69,703 +66,448 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-        SwapRequest memory swapRequest,
-        uint256 balanceTokenIn,
-        uint256 balanceTokenOut
-    ) internal virtual override whenNotPaused returns (uint256) {
-        (uint256[] memory balances, uint256 indexIn, uint256 indexOut) = _getSwapBalanceArrays(
-            swapRequest,
-            balanceTokenIn,
-            balanceTokenOut
+        (uint256 amountOut0, uint256 amountOut1) = StableMath.tokensOutFromBptIn(
+            token0.balanceOf(address(this)),
```
yuval-weiss, line 180: Why aren't we using rate-adjusted balances here?

contracts/amm/TempusAMM.sol
```diff
@@ -69,703 +66,448 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
+        view
+        override
+        returns (uint256 token0Out, uint256 token1Out)
+    {
+        // We don't need to scale balances down here
+        // as calculation for amounts out is based on btpAmountIn / totalSupply() ratio
+        // Adjusting balances with rate, and then undoing it would just cause additional calculations
+        return
+            StableMath.tokensOutFromBptIn(
+                token0.balanceOf(address(this)),
```
yuval-weiss, line 423: Why aren't we using rate-adjusted balances here?

contracts/amm/TempusAMM.sol
```diff
@@ -69,703 +66,448 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
-
-        // Exits are not completely disabled while the contract is paused: proportional exits (exact BPT in for tokens
-        // out) remain functional.
-
-        if (_isNotPaused()) {
-            // Due protocol swap fee amounts are computed by measuring the growth of the invariant between the previous
-            // join or exit event and now - the invariant's growth is due exclusively to swap fees. This avoids
-            // spending gas calculating fee amounts during each individual swap
-            dueProtocolFeeAmounts = _getDueProtocolFeeAmounts(balances, protocolSwapFeePercentage);
+    function getExpectedInGivenOut(uint256 amountOut, address tokenIn) public view override returns (uint256) {
```
yuval-weiss, line 328: can be `external`

contracts/amm/TempusAMM.sol
```diff
@@ -69,703 +66,448 @@ contract TempusAMM is BaseMinimalSwapInfoPool, StableMath, IRateProvider {
...
+        uint256 token1AmountOut,
+        uint256 maxLpTokensIn,
+        address recipient
+    ) external override whenNotPaused {
+        (uint256 balance0, uint256 balance1) = getRateAdjustedBalances();
+
+        (uint256 amountOut0, uint256 amountOut1) = getRateAdjustedAmounts(
+            token0AmountOut,
+            token1AmountOut,
+            token0.getPricePerFullShare(),
```
yuval-weiss, line 206: we have multiple consecutive `getPricePerFullShare` calls here (and in other places throughout this file). That means that we invoke `updateInterestRate` multiple times which is something we tried to avoid with the V1 implementation ([see](https://github.com/tempus-finance/tempus-protocol/blob/master/contracts/amm/TempusAMM.sol#L692))

contracts/amm/TempusAMM.sol
```diff
@@ -12,26 +12,22 @@
...
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-pragma solidity 0.7.6;
-pragma experimental ABIEncoderV2;
+pragma solidity 0.8.10;
 
-import "@balancer-labs/v2-solidity-utils/contracts/math/FixedPoint.sol";
-import "@balancer-labs/v2-solidity-utils/contracts/helpers/InputHelpers.sol";
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import "@openzeppelin/contracts/security/Pausable.sol";
```
yuval-weiss, line 18: We need to implement external functions that invoke the internal `_pause` and `_unpause` functions

contracts/TempusController.sol
```diff
@@ -251,64 +213,28 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
...
-        });
-
-        IVault.FundManagement memory fundManagement = IVault.FundManagement({
-            sender: address(this),
-            fromInternalBalance: false,
-            recipient: payable(address(this)),
-            toInternalBalance: false
-        });
-        vault.swap(singleSwap, fundManagement, minReturn, deadline);
+        tokenIn.approve(address(tempusAMM), swapAmount);
```
StefanVitor, line 221: Why are you using `approve` and not `increaseAllowance`(if there is some amount approved before)?

contracts/TempusController.sol
```diff
@@ -251,64 +213,28 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
...
-        });
-
-        IVault.FundManagement memory fundManagement = IVault.FundManagement({
-            sender: address(this),
-            fromInternalBalance: false,
-            recipient: payable(address(this)),
-            toInternalBalance: false
-        });
-        vault.swap(singleSwap, fundManagement, maxSpendAmount, deadline);
+        tokenIn.approve(address(tempusAMM), maxSpendAmount);
```
StefanVitor, line 235: Same question here (why are you using `approve` and not `increaseAllowance`)?

contracts/amm/TempusAMM.sol
mijovic: Thanks for pointing out, fixing it now

contracts/amm/TempusAMM.sol
mijovic: Fixed

contracts/amm/TempusAMM.sol
mijovic: I did add it like this in order to be able to have tokens from different pools traded. In that case there is no way to not call this twice. However, it might be that we will be targeting same underlying pools, so you are right in any case... I'll double check and update

contracts/amm/TempusAMM.sol
mijovic: https://github.com/tempus-finance/tempus-protocol/pull/576#discussion_r841848702

contracts/amm/TempusAMM.sol
mijovic: Here would be nice adding comment probably

contracts/amm/TempusAMM.sol
mijovic: I did remove this helper and made private method to be public, with `onlyOwner`

contracts/amm/TempusAMM.sol
mijovic: Unfortunately it is. `IERC20` doesn't have `decimals`, it is part of `IERC20Metadata`. If I make `IPoolShare` inherit from there it introduces an issue of implementing `decimals` in `PrincipalShare` and `YieldShare`.

mijovic: actually, since it implements it in PoolShare already we might add it to interface as well

mijovic: Done

contracts/amm/TempusAMM.sol
mijovic: Probably not

contracts/amm/TempusAMM.sol
mijovic: Reused view function so there is no doubled call to StableMath

contracts/amm/TempusAMM.sol
mijovic: Fixed it ser, wasn't that bad

contracts/amm/TempusAMM.sol
mijovic: We will add NatSpec for custom errors, so it is not worth doubling the effort

scripts/deploy.pool.config.ts
```diff
@@ -191,16 +188,12 @@ async function deployAmm(config:Config, deployerPrivateKey:string): Promise<Conf
   validateRange(config.amm.swapFee, 0, 0.03);
 
   const ammConstructorArgs = [
-    config.amm.vault,
```
StefanVitor, line 194: There is a bug in this code, which I resolved in https://github.com/tempus-finance/tempus-protocol/blob/master/scripts/deploy.pool.config.ts , so this deploy.pool.config.ts doesn't work. Maybe to fix this when we would merge with `master`, or?

mijovic: Let's fix it with merge of amm/v2 to master. That one will be painful anyway. It already has conflicts

contracts/amm/interfaces/ITempusAMM.sol
```diff
@@ -58,10 +131,7 @@ interface ITempusAMM {
     /// @param bptAmountIn amount of LP tokens in
     /// @return token0Out Amount of Token0 that user would receive back
     /// @return token1Out Amount of Token1 that user would receive back
-    function getExpectedTokensOutGivenBPTIn(uint256 bptAmountIn)
-        external
-        view
-        returns (uint256 token0Out, uint256 token1Out);
+    function getTokensOutGivenLPIn(uint256 bptAmountIn) external view returns (uint256 token0Out, uint256 token1Out);
```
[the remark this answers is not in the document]
mijovic: Fixed

StefanVitor (approved):
From my side, that's it, I would merge this PR, I think that main things are good, and for small things, we will fix when we see it in small PRs.

Issue #190, opened 2021-08-25
Switch to error types/codes from error strings


axic:
Currently error strings are extensively used, but it would be nicer to use error types.

PR #577, opened 2022-04-04
Introduce custom errors in TempusController

pashov:
Part of #190

### DIFF
--- a/contracts/ITempusController.sol
+++ b/contracts/ITempusController.sol
@@ -50,6 +50,89 @@ interface ITempusController {
         bool isEarlyRedeem
     );
 
+    /// @dev Error thrown when an unregistered contract is provided to the controller
+    /// @param deniedContract The address of the unregistered contract
+    error UnauthorizedContract(address deniedContract);
+
+    /// @dev Error thrown when an invalid leverage multiplier is provided on deposit
+    /// @param leverageMultiplier The invalid leverage multiplier amount
+    error InvalidLeverageMultiplier(uint256 leverageMultiplier);
+
+    /// @dev Error thrown when the recipient is the zero address
+    error ZeroAddressRecipient();
+
+    /// @dev Error thrown when the LP token amount is zero
+    error ZeroLPTokensAmount();
+
+    /// @dev Error thrown when the swap amount is zero
+    error ZeroSwapAmount();
+
+    /// @dev Error thrown when the maximum spend amount is zero
+    error ZeroMaxSpendAmount();
+
+    /// @dev Error thrown when the yield token amount is zero
+    error ZeroYieldTokenAmount();
+
+    /// @dev Error thrown when the backing token amount is zero
+    error ZeroBackingTokenAmount();
+
+    /// @dev Error thrown when the address of the backing token is the zero address
+    error ZeroAddressBackingToken();
+
+    /// @dev Error thrown when the address of the backing token is not the zero address
+    ///     In the case of Lido which expects deposits in Ether the code expects `backingToken = address(0)`
+    error NonZeroAddressBackingToken();
+
+    /// @dev Error thrown when the Ether value sent does not match the backing token amount provided
+    /// @param ethValue The value sent in Ether
+    /// @param backingTokenAmount The backing token amount provided
+    error EtherValueAndBackingTokenAmountMismatch(uint256 ethValue, uint256 backingTokenAmount);
+
+    /// @dev Error thrown when the principal amount and the yield amount are both zero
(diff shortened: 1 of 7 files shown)


contracts/TempusController.sol
```diff
@@ -26,16 +26,28 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
...
 
+    error UnauthorizedContract(address contractAddress);
+    error InvalidLeverageMultiplier(uint256 leverageMultiplier);
+    error ZeroAddressRecipient();
+    error ZeroLPTokensAmount();
+    error ZeroSwapAmount();
+    error ZeroMaxSpendAmount();
+    error ZeroYieldTokenAmount();
+    error ZeroBackingTokenAmount();
+    error ZeroAddressBackingToken();
```
axic, line 37: The error definitions should go into the interface (`ITempusController`).

test/utils/Utils.ts
```diff
@@ -79,6 +79,12 @@ export function getRevertMessage(e:Error): string {
   }
   let msgStart = e.message.indexOf('\'');
   if (msgStart !== -1) {
+    let braceIdx = e.message.indexOf("(");
+
+    if (braceIdx !== -1) {
+      return e.message.substring(msgStart + 1, braceIdx);
```
axic, line 85: Hm, I wonder ehetger this helper should leave this without any special marker. Currently this eould consider `revert("ZeroAddress");` and `revert ZeroAddress;`

contracts/TempusController.sol
```diff
@@ -26,16 +26,28 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
     /// Registry for valid pools and AMM's to avoid fake address injection
     mapping(address => bool) private registry;
 
+    error UnauthorizedContract(address contractAddress);
```
yuval-weiss, line 29: I think `contract` is a better name than `contractAddress` since it's of type `address` (same for other places where `contractAddress` is used throughout this file)

pashov: Yes that is what I tried but contract is a special word in Solidity and it doesn't let me use it as a variable name

test/utils/Utils.ts
pashov: I used this string manipulation, because in tests I don't always have access to the error arguments and it is hard to validate the error name and arguments, so this way I validate just the error name.

axic: I'd probably add a leading character which is not possible for custom errors, such as `:`, to make sure they are never mixed.

```suggestion
      return ":" + e.message.substring(msgStart + 1, braceIdx);
```

contracts/TempusController.sol
axic: Potentially could use a term like `target`?

test/utils/Utils.ts
axic: @StefanVitor @yuval-weiss @mijovic @RedFox20 wdyt?

pashov: IMO having the same error message as the custom error name is a rare scenario which should be stopped at code review and shouldn't need testing. I added the ":" anyway and will commit it, I am fine with both, can remove it.

contracts/ITempusController.sol
```diff
@@ -50,6 +50,89 @@ interface ITempusController {
...
+    /// @param maxSlippage The maximum slippage provided
+    error MaxSlippageTooBig(uint256 maxSlippage);
+
+    /// @dev Error thrown when maximum leftover shares are bigger or equal to both principal and yield token amounts
+    /// @param maxLeftoverShares The maximum leftover shares provided
+    error MaxLeftoverSharesTooBig(uint256 maxLeftoverShares);
+
+    /// @dev Error thrown when the Tempus AMM has not been initialized yet
+    /// @param tempusAMM The address of the Tempus AMM
+    error TempusAMMNotInitializedYet(address tempusAMM);
```
axic, line 134: Do we actually need `Tempus` in here?

pashov: Removed `Tempus`. By a quick search I found that AMM is used more than Amm, leaving it as that

test/utils/Utils.ts
RedFox20: I don't have a preference on this, looks good either way

contracts/TempusController.sol
yuval-weiss: how about `error UnauthorizedContract(address deniedContract);`?

mijovic (approved):

mijovic:
Merged because tests worked in CI

PR #586, opened 2022-04-07
amm: Refactor amplification calculation

mijovic:


axic (approved):
I think it makes not much difference, but is a bit more readable on the call sites. Did you check gas differences? Probably there should be a slight improvement.

mijovic:
I didn't check gas usage, should be improved for just a little bit. I just did this for more readability tbh

PR #588, opened 2022-04-07
amm: Fix Solidity version at 0.8.10 across the codebase

axic:
Some files are actually not locked to 0.8.x, but those are in the very minority. Not sure we should do this, but it does simplify thinking about dependencies.

### DIFF
--- a/contracts/ITempusPool.sol
+++ b/contracts/ITempusPool.sol
@@ -1,6 +1,5 @@
 // SPDX-License-Identifier: BUSL-1.1
-pragma solidity >=0.7.6 <0.9.0;
-pragma abicoder v2;
+pragma solidity 0.8.10;
 
 import "./token/IPoolShare.sol";
 import "./utils/IOwnable.sol";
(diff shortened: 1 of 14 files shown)


mijovic (approved):
I agree with locking the version. It makes it a bit harder to integrate, but it is not such a big deal. Only question is if we want interfaces to be a bit more relaxed with versioning, like >0.8?

axic:
> Only question is if we want interfaces to be a bit more relaxed with versioning, like >0.8?

Most of the interfaces depend on a given OZ version we use, so we'd need to check what is their minimum enforced version.

I also came across finding that most project don't actually integrate with pulling an interface directly, but rather having a stripped down local version, like we do in `protocols/`.

mijovic:
> > Only question is if we want interfaces to be a bit more relaxed with versioning, like >0.8?
> 
> Most of the interfaces depend on a given OZ version we use, so we'd need to check what is their minimum enforced version.
> 
> I also came across finding that most project don't actually integrate with pulling an interface directly, but rather having a stripped down local version, like we do in `protocols/`.

I would lock the version in that case

PR #593, opened 2022-04-08
feature: Tempus OTC

StefanVitor:
TODO:
- [x] NatSpec
- [x] More test for 100% TempusOTC coverage
- [x] Comments in TempusOTC typescript class
- [x] Merge with latest changes from `master`

### DIFF
--- a/contracts/otc/ITempusOTC.sol
+++ b/contracts/otc/ITempusOTC.sol
@@ -1,0 +1,163 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.10;
+
+import "../ITempusPool.sol";
+
+/// @dev OTC (Over The Counter) trading is directly trade between two parties, where they agree on a price
+///     and then work out the transfer of assets between themselves.
+/// In this case, steps are next:
+///     -   The offer setter (the side that wants to buy Yields) deploys `TempusOTC` contract
+///             and creates an offer with the following information:
+///         -   How many Principals he wants to sell
+///         -   How many Yields is requested from the other side to buy that amount of Principals
+///     -   The other side accepts an offer, provides Yields that the offer setter requested, and gets Principals
+///     -   At any time, the offer setter may cancel the offer and get his Principals back
+interface ITempusOTC {
+    /// @dev Event emitted on a successful BT/YBT deposit.
+    /// @param pool The Tempus Pool to which assets were deposited
+    /// @param depositor Address of the user who deposited Yield Bearing Tokens to mint
+    ///                  Tempus Principal Share (TPS) and Tempus Yield Shares
+    /// @param recipient Address of the recipient who will receive TPS and TYS tokens
+    /// @param yieldTokenAmount Amount of yield tokens received from underlying pool
+    /// @param backingTokenValue Value of @param yieldTokenAmount expressed in backing tokens
+    /// @param shareAmounts Number of Tempus Principal Shares (TPS) and Tempus Yield Shares (TYS) granted to `recipient`
+    /// @param interestRate Interest Rate of the underlying pool from Yield Bearing Tokens to the underlying asset
+    /// @param fee The fee which was deducted (in terms of yield bearing tokens)
+    event DepositedOTC(
+        ITempusPool indexed pool,
+        address indexed depositor,
+        address indexed recipient,
+        uint256 yieldTokenAmount,
+        uint256 backingTokenValue,
+        uint256 shareAmounts,
+        uint256 interestRate,
+        uint256 fee
+    );
+
+    /// @dev Event emitted on a successful BT/YBT redemption.
+    /// @param pool The Tempus Pool from which Tempus Shares were redeemed
+    /// @param redeemer Address of the user whose Shares (Principals and Yields) are redeemed
+    /// @param recipient Address of user that received Yield Bearing Tokens
+    /// @param principalShareAmount Number of Tempus Principal Shares (TPS) to redeem into the Yield Bearing Token (YBT)
+    /// @param yieldShareAmount Number of Tempus Yield Shares (TYS) to redeem into the Yield Bearing Token (YBT)
+    /// @param yieldTokenAmount Number of Yield bearing tokens redeemed from the pool
+    /// @param backingTokenValue Value of @param yieldTokenAmount expressed in backing tokens
+    /// @param interestRate Interest Rate of the underlying pool from Yield Bearing Tokens to the underlying asset
+    /// @param fee The fee which was deducted (in terms of yield bearing tokens)
+    /// @param isEarlyRedeem True in case of early redemption, otherwise false
+    event RedeemedOTC(
+        ITempusPool indexed pool,
+        address indexed redeemer,
+        address indexed recipient,
+        uint256 principalShareAmount,
+        uint256 yieldShareAmount,
+        uint256 yieldTokenAmount,
+        uint256 backingTokenValue,
+        uint256 interestRate,
+        uint256 fee,
+        bool isEarlyRedeem
+    );
+
+    /// @dev Error when setting offer which was already set
+    error OfferAlreadySet();
+
+    /// @dev Error when accepting non-existant offer
+    error OfferNotCreated();
+
+    /// @dev Error when offer is not accepted
+    error OfferNotAccepted();
+
+    /// @dev Error when msg.sender is not YieldReceiver during yield withdraw
+    error YieldReceiverIsNotSameAsMsgSender();
+
+    /// @dev Error when there is no enough yield shares to withdraw
+    error NoEnoughYieldSharesToWithdraw();
+
+    /// @dev Error when there is no enough shares to accept offer
+    error NoEnoughSharesToAcceptOffer();
+
+    /// @dev Error thrown when the principal amount and the yield amount are both zero
+    error ZeroPrincipalAndYieldAmounts();
+
+    /// @dev Error thrown when the yield token amount is zero
+    error ZeroYieldTokenAmount();
+
+    /// @dev Error thrown when the backing token amount is zero
+    error ZeroBackingTokenAmount();
+
+    /// @dev Error thrown when the address of the backing token is the zero address
+    error ZeroAddressBackingToken();
+
+    /// @dev Error thrown when the address of the backing token is not the zero address
+    ///     In the case of Lido which expects deposits in Ether the code expects `backingToken = address(0)`
+    error NonZeroAddressBackingToken();
+
+    /// @dev Error thrown when the Ether value sent does not match the backing token amount provided
+    /// @param ethValue The value sent in Ether
+    /// @param backingTokenAmount The backing token amount provided
+    error EtherValueAndBackingTokenAmountMismatch(uint256 ethValue, uint256 backingTokenAmount);
+
+    /// @dev In which state is offer
+    ///     NotSet Offer is in the initial phase (it's not set, or was set, but after that was cancelled)
+    ///     Created Offer is set
+    ///     Accepted Offer is accepted
+    enum OfferStatus {
+        NotSet,
+        Created,
+        Accepted
+    }
+
+    /// @dev Create offer in next steps:
+    ///     Mint TPS (Tempus Pool Shares) and TYS (Tempus Yield Shares)
+    ///         from YBT (Yield Bearing Token) or BT (Backing Token)
+    ///     Set TPS for sell(offer) and get TYS to `recipient`
+    /// @param tokenAmount YBT/BT amount which will be deposited to get yield and principals
+    /// @param isBackingToken Specifies whether the deposited asset is the Yield Bearing Token or Backing Token
+    /// @param requestYieldAmount Amount in TYS which user that accepts an offer should provide
+    /// @param recipient Address of the recipient who will receive TYS tokens (immediately and after offer accepted)
+    function setOffer(
+        uint256 tokenAmount,
+        bool isBackingToken,
+        uint256 requestYieldAmount,
+        address recipient
+    ) external payable returns (uint256);
+
+    /// @dev Cancel offer that is created and get principals(TPS) that are set for sell (created in setOffer)
+    /// @param recipient Address of the recipient who will receive TPS tokens that are created in setOffer
+    function cancelOffer(address recipient) external returns (uint256);
+
+    /// @dev Accept offer that is created with setOffer in next steps:
+    ///     Mint TPS (Tempus Pool Shares) and TYS (Tempus Yield Shares)
+    ///         from YBT (Yield Bearing Token) or BT (Backing Token)
+    ///     Get TPS that are set for offer(+minted) and minted TYS which remained from requested amount
+    ///     Must be equal or more shares that offer setter requested
+    /// @param tokenAmount YBT/BT amount which will be deposited to get yield and principals
+    /// @param isBackingToken Specifies whether the deposited asset is the Yield Bearing Token or Backing Token
+    /// @param recipient Address of the recipient who will receive TPS and TYS
+    /// @return principalAmount Get TPS amount which is sum minted amount and set offer amount
+    /// @return yieldAmount Get TYS amount which is difference minted amount and yield requested amount
+    function acceptOffer(
+        uint256 tokenAmount,
+        bool isBackingToken,
+        address recipient
+    ) external payable returns (uint256 principalAmount, uint256 yieldAmount);
+
+    /// @dev Withdraw TYS (Tempus Yield Shares) after offer accepted
+    /// @param tokenAmount TYS amount which is withdraw
+    function withdrawYieldAfterOfferAccepted(uint256 tokenAmount) external;
+
+    /// @dev Redeem TPS+TYS held by msg.sender into BT(Backing Tokens) or YBT(Yield Bearing Tokens)
+    /// @notice `msg.sender` will receive backing tokens or yield bearing tokens
+    /// @notice Before maturity, `principalAmount` must equal to `yieldAmount`
+    /// @param principals Amount of Tempus Principals to redeem in PrincipalShare decimal precision
+    /// @param yields Amount of Tempus Yields to redeem in YieldShare decimal precision
+    /// @param recipient Address of user that will receive BT/YBT
+    /// @param isBackingToken Specifies whether the reedem asset will be in Backing Token or Yield Bearing Token
+    /// @return Amount of BT/YBT that were imbursed as a result of the redemption
+    function redeem(
+        uint256 principals,
+        uint256 yields,
+        address recipient,
+        bool isBackingToken
+    ) external returns (uint256);
+}
--- a/contracts/otc/TempusOTC.sol
+++ b/contracts/otc/TempusOTC.sol
@@ -1,0 +1,272 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.10;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+
+import "./ITempusOTC.sol";
+import "../ITempusPool.sol";
+import "../utils/Ownable.sol";
+import "../utils/UntrustedERC20.sol";
+
+contract TempusOTC is ITempusOTC, ReentrancyGuard, Ownable {
+    using UntrustedERC20 for IERC20;
+
+    ITempusPool public immutable tempusPool;
+
+    OfferStatus public offerStatus;
+    address public yieldReceiver;
+    uint256 public principalSetOfferAmount;
+    uint256 public yieldRequestedAmount;
+
+    constructor(ITempusPool pool) {
+        tempusPool = pool;
+        offerStatus = OfferStatus.NotSet;
+    }
+
+    /// @inheritdoc ITempusOTC
+    function setOffer(
+        uint256 tokenAmount,
+        bool isBackingToken,
+        uint256 requestYieldAmount,
+        address recipient
+    ) external payable onlyOwner returns (uint256) {
+        if (offerStatus != OfferStatus.NotSet) {
+            revert OfferAlreadySet();
+        }
+
+        uint256 mintedShares = isBackingToken ? _depositBacking(tokenAmount) : _depositYieldBearing(tokenAmount);
+
+        _fillOfferDetails(recipient, OfferStatus.Created, mintedShares, requestYieldAmount);
+
+        IERC20 yieldShares = IERC20(address(tempusPool.yieldShare()));
+        yieldShares.transfer(recipient, mintedShares);
+
+        return mintedShares;
+    }
+
+    /// @inheritdoc ITempusOTC
+    function cancelOffer(address recipient) external onlyOwner returns (uint256) {
+        if (offerStatus != OfferStatus.Created) {
+            revert OfferNotCreated();
+        }
+
+        uint256 principalLockedAmount = principalSetOfferAmount;
+
+        _fillOfferDetails(address(0), OfferStatus.NotSet, 0, 0);
+
+        IERC20 principalShares = IERC20(address(tempusPool.principalShare()));
+        principalShares.transfer(recipient, principalLockedAmount);
+
+        return principalLockedAmount;
+    }
+
+    /// @inheritdoc ITempusOTC
+    function acceptOffer(
+        uint256 tokenAmount,
+        bool isBackingToken,
+        address recipient
+    ) external payable nonReentrant returns (uint256 principalAmount, uint256 yieldAmount) {
+        if (offerStatus != OfferStatus.Created) {
+            revert OfferNotCreated();
+        }
+
+        if (tempusPool.getSharesAmountForExactTokensOut(tokenAmount, isBackingToken) < yieldRequestedAmount) {
+            revert NoEnoughSharesToAcceptOffer();
+        }
+
+        uint256 mintedShares = isBackingToken ? _depositBacking(tokenAmount) : _depositYieldBearing(tokenAmount);
+
+        offerStatus = OfferStatus.Accepted;
+
+        principalAmount = mintedShares + principalSetOfferAmount;
+        IERC20 principalShares = IERC20(address(tempusPool.principalShare()));
+        principalShares.transfer(recipient, principalAmount);
+
+        yieldAmount = mintedShares - yieldRequestedAmount;
+        IERC20 yieldShares = IERC20(address(tempusPool.yieldShare()));
+        yieldShares.transfer(recipient, yieldAmount);
+    }
+
+    /// @inheritdoc ITempusOTC
+    function withdrawYieldAfterOfferAccepted(uint256 tokenAmount) external {
+        if (offerStatus != OfferStatus.Accepted) {
+            revert OfferNotAccepted();
+        }
+
+        if (msg.sender != yieldReceiver) {
+            revert YieldReceiverIsNotSameAsMsgSender();
+        }
+
+        IERC20 yieldShares = IERC20(address(tempusPool.yieldShare()));
+        if (tokenAmount > yieldShares.balanceOf(address(this))) {
+            revert NoEnoughYieldSharesToWithdraw();
+        }
+
+        yieldShares.transfer(msg.sender, tokenAmount);
+    }
+
+    /// @inheritdoc ITempusOTC
+    function redeem(
+        uint256 principals,
+        uint256 yields,
+        address recipient,
+        bool isBackingToken
+    ) external nonReentrant returns (uint256) {
+        if (principals == 0 && yields == 0) {
+            revert ZeroPrincipalAndYieldAmounts();
+        }
+
+        return
+            isBackingToken
+                ? _redeemToBacking(msg.sender, principals, yields, recipient)
+                : _redeemToYieldBearing(msg.sender, principals, yields, recipient);
+    }
+
+    function _depositYieldBearing(uint256 yieldTokenAmount) private returns (uint256) {
+        if (yieldTokenAmount == 0) {
+            revert ZeroYieldTokenAmount();
+        }
+
+        IERC20 yieldBearingToken = IERC20(tempusPool.yieldBearingToken());
+
+        // Transfer funds from msg.sender to tempusPool
+        uint256 transferredYBT = yieldBearingToken.untrustedTransferFrom(
+            msg.sender,
+            address(tempusPool),
+            yieldTokenAmount
+        );
+
+        (uint256 mintedShares, uint256 depositedBT, uint256 fee, uint256 rate) = tempusPool.onDepositYieldBearing(
+            transferredYBT,
+            address(this)
+        );
+
+        emit DepositedOTC(tempusPool, msg.sender, address(this), transferredYBT, depositedBT, mintedShares, rate, fee);
+
+        return mintedShares;
+    }
+
+    function _depositBacking(uint256 _backingTokenAmount) private returns (uint256) {
+        if (_backingTokenAmount == 0) {
+            revert ZeroBackingTokenAmount();
+        }
+
+        IERC20 backingToken = IERC20(tempusPool.backingToken());
+
+        // In case the underlying pool expects deposits in Ether (e.g. Lido),
+        // it uses `backingToken = address(0)`.  Since we disallow 0-value deposits,
+        // and `msg.value == backingTokenAmount`, this check here can be used to
+        // distinguish between the two pool types.
+        if (msg.value == 0) {
+            // NOTE: We need to have this check here to avoid calling transfer on address(0),
+            //       because that always succeeds.
+            if (address(backingToken) == address(0)) {
+                revert ZeroAddressBackingToken();
+            }
+
+            _backingTokenAmount = backingToken.untrustedTransferFrom(
+                msg.sender,
+                address(tempusPool),
+                _backingTokenAmount
+            );
+        } else {
+            if (address(backingToken) != address(0)) {
+                revert NonZeroAddressBackingToken();
+            }
+            if (msg.value != _backingTokenAmount) {
+                revert EtherValueAndBackingTokenAmountMismatch(msg.value, _backingTokenAmount);
+            }
+        }
+
+        (uint256 mintedShares, uint256 depositedYBT, uint256 fee, uint256 interestRate) = tempusPool.onDepositBacking{
+            value: msg.value
+        }(_backingTokenAmount, address(this));
+
+        emit DepositedOTC(
+            tempusPool,
+            msg.sender,
+            address(this),
+            depositedYBT,
+            _backingTokenAmount,
+            mintedShares,
+            interestRate,
+            fee
+        );
+
+        return mintedShares;
+    }
+
+    function _redeemToYieldBearing(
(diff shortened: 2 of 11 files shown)


contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,358 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.10;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+
+import "../TempusPool.sol";
```
axic, line 7: Nit, should only use the interfaces (`ITempusPool)` unless necessary otherwise.

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,358 @@
...
+
+contract TempusOTC is ReentrancyGuard, Ownable {
+    using UntrustedERC20 for IERC20;
+
+    TempusPool public tempusPool;
```
axic, line 12: Should be immutable.

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,358 @@
...
+    /// @param pool The Tempus Pool to which assets were deposited
+    /// @param depositor Address of the user who deposited Yield Bearing Tokens to mint
+    ///                  Tempus Principal Share (TPS) and Tempus Yield Shares
+    /// @param recipient Address of the recipient who will receive TPS and TYS tokens
+    /// @param yieldTokenAmount Amount of yield tokens received from underlying pool
+    /// @param backingTokenValue Value of @param yieldTokenAmount expressed in backing tokens
+    /// @param shareAmounts Number of Tempus Principal Shares (TPS) and Tempus Yield Shares (TYS) granted to `recipient`
+    /// @param interestRate Interest Rate of the underlying pool from Yield Bearing Tokens to the underlying asset
+    /// @param fee The fee which was deducted (in terms of yield bearing tokens)
+    event DepositedOTC(
```
axic, line 29: All the events, errors, structs, etc. should go into an interface, `ITempusOTC`.

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,358 @@
...
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+
+import "../TempusPool.sol";
+
+contract TempusOTC is ReentrancyGuard, Ownable {
+    using UntrustedERC20 for IERC20;
+
+    TempusPool public tempusPool;
+
+    OfferStatus public offerStatus;
```
axic, line 14: I'm confused, is this contract per deal? Otherwise it should have a mapping for the orderbook.

axic (commented):
Could have a look at "EtherDelta" or any other on-chain orderbook implementation for some inspiration.

Also wonder how this could/would work together with #554?

contracts/otc/TempusOTC.sol
StefanVitor: Resolved

contracts/otc/TempusOTC.sol
StefanVitor: Resolved

contracts/otc/TempusOTC.sol
StefanVitor: Resolved

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,271 @@
...
+        uint256 tokenAmount,
+        bool isBackingToken,
+        uint256 requestYieldAmount,
+        address recipient
+    ) external payable onlyOwner {
+        if (offerStatus != OfferStatus.NotSet) {
+            revert OfferStatusMustBeInNotSetStatus();
+        }
+
+        uint256 mintedShares = isBackingToken ? _depositBacking(tokenAmount) : _depositYieldBearing(tokenAmount);
```
axic, line 37: Cannot this function use `controller.depositBacking`/`depositYieldBearing`?

If not, can we think about how to do that to reduce code duplication?

StefanVitor: No, because the TempusPool controller is TempusOTC and not TempusController. And if you use TempusController, then the offer creator should deploy and TempusController, and that's expensive and not practic. As you said, maybe we should try to find a solution to reduce code duplication.

axic: > TempusPool controller is TempusOTC

Okay I think there should be more context for the reviewers. This is confusing to me 😅

StefanVitor: "TempusPool controller" is immutable variable `controller` in contract `TempusPool`. That immutable variable is checks in modifier `onlyController`, and because that variable is set to `TempusOTC` address in a constructor, it's not possible to use a controller.depositBacking/depositYieldBearing (so, in this case, "TempusPool controller will be TempusOTC"). On the other side, if we initialize the `controller` immutable variable to the `TempusController` contract, that means that every offer setter would deploy and `TempusController` contract, and that is, in my opinion, expensive, and maybe could provide additional problems.

contracts/otc/ITempusOTC.sol
```diff
@@ -0,0 +1,152 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.10;
+
+interface ITempusOTC {
```
RedFox20, line 4: Interface is missing a brief documentation. We must assume that this will be viewed by wider community who have no idea what over the counter trading is, so this should described in brief

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,276 @@
...
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+
+import "./ITempusOTC.sol";
+import "../ITempusPool.sol";
+import "../utils/Ownable.sol";
+import "../utils/UntrustedERC20.sol";
+
+contract TempusOTC is ITempusOTC, ReentrancyGuard, Ownable {
```
RedFox20, line 12: And this should be marked with `@inheritdoc`

contracts/otc/ITempusOTC.sol
```diff
@@ -0,0 +1,152 @@
...
+    /// @param recipient Address of user that received Yield Bearing Tokens
+    /// @param principalShareAmount Number of Tempus Principal Shares (TPS) to redeem into the Yield Bearing Token (YBT)
+    /// @param yieldShareAmount Number of Tempus Yield Shares (TYS) to redeem into the Yield Bearing Token (YBT)
+    /// @param yieldTokenAmount Number of Yield bearing tokens redeemed from the pool
+    /// @param backingTokenValue Value of @param yieldTokenAmount expressed in backing tokens
+    /// @param interestRate Interest Rate of the underlying pool from Yield Bearing Tokens to the underlying asset
+    /// @param fee The fee which was deducted (in terms of yield bearing tokens)
+    /// @param isEarlyRedeem True in case of early redemption, otherwise false
+    event RedeemedOTC(
+        address indexed pool,
```
RedFox20, line 38: If we used `ITempusPool indexed pool` here, could we perhaps avoid `address(tempusPool)` casts?

test/otc/TempusOTC.test.ts
```diff
@@ -0,0 +1,262 @@
...
+  async function acceptOfferDefault() {
+    await tempusOTC.acceptOffer(testPool, acceptOfferUser, acceptOfferSendAmount, false);
+  }
+
+  async function withdrawYieldAfterOfferAcceptedDefault() {
+    await tempusOTC.withdrawYieldAfterOfferAccepted(testPool, setOfferUser, yieldRequestedAmount);
+  }
+
+  async function setOfferChecks() {
+    expect(await pool.yieldShare.balanceOf(tempusOTC.address)).to.equal(0);
```
RedFox20, line 49: `tempusOTC.address` is redundant here I think, unless you want it to read like this which is also okay I guess

test/otc/TempusOTC.test.ts
```diff
@@ -0,0 +1,262 @@
...
+    (await expectRevert(tempusOTC.acceptOffer(testPool, acceptOfferUser, 1, false))).to.equal(':NoEnoughSharesToAcceptOffer');
+
+    await acceptOfferDefault();
+    (await expectRevert(acceptOfferDefault())).to.equal(':OfferStatusMustBeInCreatedStatus');
+  });
+
+  it("check if [withdrawYieldAfterOfferAccepted] correct", async () =>
+  {
+    await setOfferDefault();
+
```
RedFox20, line 172: Probably the newline between each function call is a bit redundant and not mandatory 😄

test/otc/TempusOTC.test.ts
```diff
@@ -0,0 +1,262 @@
...
+    await withdrawYieldAfterOfferAcceptedDefault();
+
+    await testPool.fastForwardToMaturity();
+
+    const yieldBearingBeforeRedeem = await pool.yieldBearing.balanceOf(acceptOfferUser.address);
+    const principalShareAmount = await pool.principalShare.balanceOf(acceptOfferUser.address);
+    const yieldShareAmount = await pool.yieldShare.balanceOf(acceptOfferUser.address);
+
+    await tempusOTC.redeem(
```
RedFox20, line 210: This should fit on one line, inside our current 120col limit

test/pool-utils/AaveTestPool.ts
```diff
@@ -32,4 +32,13 @@ export class AaveTestPool extends PoolTestFixture {
       this.ybt = this.aave.yieldToken;
     });
   }
+  async createPoolForOTC(params:TempusParams): Promise<TempusPool> {
+    return await this.initPoolForOTC(params, this.YIELD_TOKEN.name, this.YIELD_TOKEN.symbol, async () => {
+      return await Aave.create(this.ASSET_TOKEN, this.YIELD_TOKEN, this.initialRate);
+    }, (pool:ContractBase) => {
+      this.aave = <Aave>pool;
+      this.asset = this.aave.asset;
+      this.ybt = this.aave.yieldToken;
```
RedFox20, line 41: This part should not be duplicated. If you could find a way to not duplicate `Aave.create` and these variable inits, it would be much better.

test/pool-utils/LidoTestPool.ts
```diff
@@ -54,4 +54,17 @@ export class LidoTestPool extends PoolTestFixture {
...
+    return await this.initPoolForOTC(params, this.YIELD_TOKEN.name, this.YIELD_TOKEN.symbol, async () => {
+      if (this.integration) {
+        return await LidoFork.create(this.ASSET_TOKEN, this.YIELD_TOKEN, this.initialRate);
+      } else {
+        return await LidoMock.create(this.ASSET_TOKEN, this.YIELD_TOKEN, this.initialRate);
+      }
+    }, (pool:ContractBase) => {
+      this.lido = <LidoContract>pool;
+      this.asset = this.lido.asset;
+      this.ybt = this.lido.yieldToken;
```
RedFox20, line 67: Same here, this initialization logic should not be duplicated like this 🙂

test/pool-utils/PoolTestFixture.ts
```diff
@@ -502,4 +518,68 @@ export abstract class PoolTestFixture {
     this.yields = this.tempus.yieldShare;
     return this.tempus;
   }
+
+  protected async initPoolForOTC(
```
RedFox20, line 522: Although there are unique differences in this, it is also a huge duplication of initialization again, which could be solved by using a few specific callbacks.

The main problem is, imagine we have a maintenance task where some of the initialization needs to be changed, but developer forgets to fix the second duplicated part... Duplication is a huge burden as the project gets older.

contracts/otc/ITempusOTC.sol
```diff
@@ -0,0 +1,152 @@
...
+        uint256 yieldShareAmount,
+        uint256 yieldTokenAmount,
+        uint256 backingTokenValue,
+        uint256 interestRate,
+        uint256 fee,
+        bool isEarlyRedeem
+    );
+
+    /// @dev Error when offerStatus must be in NotSet status
+    error OfferStatusMustBeInNotSetStatus();
```
yuval-weiss, line 51:
```suggestion
    error OfferAlreadySet();
```

contracts/otc/ITempusOTC.sol
```diff
@@ -0,0 +1,152 @@
...
+
+    /// @dev Error when offerStatus must be in Created status
+    error OfferStatusMustBeInCreatedStatus();
```
yuval-weiss, line 54:
```suggestion
    error OfferNotCreated();
```

test/utils/TempusOTC.ts
```diff
@@ -0,0 +1,126 @@
...
+    yieldRequestedAmount:NumberOrString, 
+    ethValue: NumberOrString = undefined,
+    recipient: SignerOrAddress = user
+  ): Promise<Transaction>{
+    await this.approve(pool, user, tokenAmount, isBackingToken);
+    const t = pool.tempus;
+    const depositAmount = isBackingToken ? t.asset.toBigNum(tokenAmount) : t.yieldBearing.toBigNum(tokenAmount);
+    const ethToTransfer = pool.acceptsEther ? ((ethValue == undefined) ? tokenAmount : ethValue) : (ethValue || 0);
+    
+    return this.contract.connect(user).setOffer(
```
RedFox20, line 60: `this.connect(user)....` should also work

contracts/otc/TempusOTC.sol
axic: I meant the lack of detail about a high level overview of the flow. That could also go into the natspec for the interface.

contracts/otc/ITempusOTC.sol
StefanVitor: Now I add an explanation, check that, and write if you think that could be more detailed.

contracts/otc/TempusOTC.sol
StefanVitor: It's not possible to use @inheritdoc with the contract, it could be only function by function (`context - function, public state variable`)- https://docs.soliditylang.org/en/v0.8.13/natspec-format.html?highlight=inheritdoc#documentation-example and I add that.

contracts/otc/ITempusOTC.sol
StefanVitor: Resolved

contracts/otc/ITempusOTC.sol
```diff
@@ -0,0 +1,164 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.10;
+
+import "../ITempusPool.sol";
+
+/// @dev OTC (Over The Counter) trading is directly trade between two parties, where they agree on a price
+///     and then work out the transfer of assets between themselves.
+/// In this case, steps are next:
+///     -   The offer setter (the side that wants to sell an asset) deploys TempusOTC contract
```
axic, line 9: I'd actually make the `TempusOTC` contract into a factory, which creates the instance where each of the properties are immutable.

test/otc/TempusOTC.test.ts
StefanVitor: For users is redundant, `tempusOTC.address` must write.

test/pool-utils/LidoTestPool.ts
StefanVitor: Same as previous - "I implement other solution for this, check that and write your opinion (with argument isOTC in createWithAMM function)"

contracts/otc/ITempusOTC.sol
StefanVitor: Resolved

test/utils/TempusOTC.ts
StefanVitor: Resolved

contracts/otc/ITempusOTC.sol
StefanVitor: Resolved

contracts/otc/ITempusOTC.sol
mijovic, line 9:
```suggestion
///     -   The offer setter (the side that wants to buy Yields) deploys `TempusOTC` contract
```

contracts/otc/ITempusOTC.sol
```diff
@@ -0,0 +1,164 @@
...
+///             and creates an offer with the next information:
```
mijovic, line 10:
```suggestion
///             and creates an offer with the following information:
```

contracts/otc/ITempusOTC.sol
```diff
@@ -0,0 +1,164 @@
...
+///         -   How much Tempus Principal Share (TPS) wants to sell
```
mijovic, line 11:
```suggestion
///         -   How many Principals wants to sell
```

contracts/otc/ITempusOTC.sol
```diff
@@ -0,0 +1,164 @@
...
+///         -   How much Tempus Yield Share (TYS) request from the other side to buy that TPS
```
mijovic, line 12:
```suggestion
///         -   How many Yields is requested from the other side to buy that amount of Principals
```

contracts/otc/ITempusOTC.sol
```diff
@@ -0,0 +1,164 @@
...
+///     -   The other side accepts an offer, provides TYS that offer setter request, and gets TPS
```
mijovic, line 13:
```suggestion
///     -   The other side accepts an offer, provides Yields that offer setter requested, and gets Principals
```

contracts/otc/ITempusOTC.sol
```diff
@@ -0,0 +1,164 @@
...
+///     -   In the meantime, in every time, the offer setter may cancel the offer
```
mijovic, line 14:
```suggestion
///     -   At any time, the offer setter may cancel the offer and get his Principals back
```

contracts/otc/ITempusOTC.sol
```diff
@@ -0,0 +1,164 @@
...
+///     -   In every time, it's possible to redeem TPS+TYS and get BT(Backing Tokens) or YBT(Yield Bearing Tokens)
```
mijovic, line 15: I think this is irrelevant here

contracts/otc/ITempusOTC.sol
```diff
@@ -0,0 +1,163 @@
...
+        uint256 backingTokenValue,
+        uint256 interestRate,
+        uint256 fee,
+        bool isEarlyRedeem
+    );
+
+    /// @dev Error when offerStatus must be in NotSet status
+    error OfferAlreadySet();
+
+    /// @dev Error when offerStatus must be in Created status
```
mijovic, line 64: How about?
```suggestion
    /// @dev Error when accepting non-existant offer
```

contracts/otc/ITempusOTC.sol
```diff
@@ -0,0 +1,163 @@
...
+        uint256 principalShareAmount,
+        uint256 yieldShareAmount,
+        uint256 yieldTokenAmount,
+        uint256 backingTokenValue,
+        uint256 interestRate,
+        uint256 fee,
+        bool isEarlyRedeem
+    );
+
+    /// @dev Error when offerStatus must be in NotSet status
```
mijovic, line 61:
```suggestion
    /// @dev Error when setting offer which was already set
```

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+        _fillOfferDetails(recipient, OfferStatus.Created, mintedShares, requestYieldAmount);
+
+        IERC20 yieldShares = IERC20(address(tempusPool.yieldShare()));
+        yieldShares.transfer(recipient, mintedShares);
+
+        return mintedShares;
+    }
+
+    /// @inheritdoc ITempusOTC
+    function cancelOffer(address recipient) external onlyOwner returns (uint256) {
```
yuval-weiss, line 49:
```suggestion
    function cancelOffer(address recipient) external onlyOwner nonReentrant returns (uint256) {
```

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+        offerStatus = OfferStatus.NotSet;
+    }
+
+    /// @inheritdoc ITempusOTC
+    function setOffer(
+        uint256 tokenAmount,
+        bool isBackingToken,
+        uint256 requestYieldAmount,
+        address recipient
+    ) external payable onlyOwner returns (uint256) {
```
yuval-weiss, line 33:
```suggestion
    ) external payable onlyOwner nonReentrant returns (uint256) {
```

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+        IERC20 principalShares = IERC20(address(tempusPool.principalShare()));
+        principalShares.transfer(recipient, principalAmount);
+
+        yieldAmount = mintedShares - yieldRequestedAmount;
+        IERC20 yieldShares = IERC20(address(tempusPool.yieldShare()));
+        yieldShares.transfer(recipient, yieldAmount);
+    }
+
+    /// @inheritdoc ITempusOTC
+    function withdrawYieldAfterOfferAccepted(uint256 tokenAmount) external {
```
yuval-weiss, line 92: Why not just transfer the entire amount of yields to the Taker in `acceptOffer`?

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+        return principalLockedAmount;
+    }
+
+    /// @inheritdoc ITempusOTC
+    function acceptOffer(
+        uint256 tokenAmount,
+        bool isBackingToken,
+        address recipient
+    ) external payable nonReentrant returns (uint256 principalAmount, uint256 yieldAmount) {
+        if (offerStatus != OfferStatus.Created) {
```
yuval-weiss, line 70: A `msg.sender == yieldReceiver` check is missing here.

contracts/otc/TempusOTC.sol
StefanVitor: I thought to add this, but in my opinion, it's unnecessary because it has the `onlyOwner` modifier, so it means that the contract owner will potentially "hack" his own contract. And also, shares from the other side(from the side that accepts the offer) are not stored in the contract, so the contract owner could potentially hack only his own tokens. What do other guys think about this (for me, it's no problem to add `nonReentrant` if you think that it is necessary)

contracts/otc/TempusOTC.sol
StefanVitor: I also thought about this, but in my opinion, that is not in good solidity programming style because one user will pay gas for a transaction for the other user, and also, the other user will not have a possibility to withdraw that amount when he wants (and how much he want, he will only automatically get tokens). I prefer this style, but, as I write for the previous message, what do other guys think about this?

contracts/otc/TempusOTC.sol
StefanVitor: No, maybe `yieldReceiver` is not a good named, because that is `setOfferReceiver` (perhaps that is the better naming). So, everyone could accept the offer, but maybe, I could add that in `setOffer`, one parameter would be an address that could accept the offer (as I said, now, everyone could accept, with that, offer setter will be set who could accept offer).

contracts/otc/TempusOTC.sol
yuval-weiss: Yeah it makes sense considering the fact that the design allows anyone to acceptOffer...

contracts/otc/TempusOTC.sol
yuval-weiss: I think it's alright as is

contracts/otc/ITempusOTC.sol
```diff
@@ -0,0 +1,163 @@
...
+
+/// @dev OTC (Over The Counter) trading is directly trade between two parties, where they agree on a price
+///     and then work out the transfer of assets between themselves.
+/// In this case, steps are next:
+///     -   The offer setter (the side that wants to buy Yields) deploys `TempusOTC` contract
+///             and creates an offer with the following information:
+///         -   How many Principals he wants to sell
+///         -   How many Yields is requested from the other side to buy that amount of Principals
+///     -   The other side accepts an offer, provides Yields that the offer setter requested, and gets Principals
+///     -   At any time, the offer setter may cancel the offer and get his Principals back
```
RedFox20, line 14: Sorry to nitpick, but there are some English typos here, but I think our public-facing docs should be flawless to give a great first impression. I suggest you use an advanced English spellchecker such as Grammarly, it has saved our team a huge amount of time with this :)

> @dev OTC (Over The Counter) ~~trading is directly trade~~ between two parties, where they agree on a price
"is direct trading"
> In this case, steps are next:
"The offer steps are as follows:"

That should be all that I noticed right now 👍🏻

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+    function cancelOffer(address recipient) external onlyOwner returns (uint256) {
+        if (offerStatus != OfferStatus.Created) {
+            revert OfferNotCreated();
+        }
+
+        uint256 principalLockedAmount = principalSetOfferAmount;
+
+        _fillOfferDetails(address(0), OfferStatus.NotSet, 0, 0);
+
+        IERC20 principalShares = IERC20(address(tempusPool.principalShare()));
```
RedFox20, line 58: This cast is not needed if you rebase to master

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+    ) external payable onlyOwner returns (uint256) {
+        if (offerStatus != OfferStatus.NotSet) {
+            revert OfferAlreadySet();
+        }
+
+        uint256 mintedShares = isBackingToken ? _depositBacking(tokenAmount) : _depositYieldBearing(tokenAmount);
+
+        _fillOfferDetails(recipient, OfferStatus.Created, mintedShares, requestYieldAmount);
+
+        IERC20 yieldShares = IERC20(address(tempusPool.yieldShare()));
```
RedFox20, line 42: This cast is not needed if you rebase to master

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+        if (tempusPool.getSharesAmountForExactTokensOut(tokenAmount, isBackingToken) < yieldRequestedAmount) {
+            revert NoEnoughSharesToAcceptOffer();
+        }
+
+        uint256 mintedShares = isBackingToken ? _depositBacking(tokenAmount) : _depositYieldBearing(tokenAmount);
+
+        offerStatus = OfferStatus.Accepted;
+
+        principalAmount = mintedShares + principalSetOfferAmount;
+        IERC20 principalShares = IERC20(address(tempusPool.principalShare()));
```
RedFox20, line 83: This cast is not needed if you rebase to master

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+        principalShares.transfer(recipient, principalAmount);
+
+        yieldAmount = mintedShares - yieldRequestedAmount;
+        IERC20 yieldShares = IERC20(address(tempusPool.yieldShare()));
```
RedFox20, line 87: Also here, cast no longer needed

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+    function withdrawYieldAfterOfferAccepted(uint256 tokenAmount) external {
+        if (offerStatus != OfferStatus.Accepted) {
+            revert OfferNotAccepted();
+        }
+
+        if (msg.sender != yieldReceiver) {
+            revert YieldReceiverIsNotSameAsMsgSender();
+        }
+
+        IERC20 yieldShares = IERC20(address(tempusPool.yieldShare()));
```
RedFox20, line 101: This cast is not needed if you rebase to master

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+        bool isBackingToken
+    ) external nonReentrant returns (uint256) {
+        if (principals == 0 && yields == 0) {
+            revert ZeroPrincipalAndYieldAmounts();
+        }
+
+        return
+            isBackingToken
+                ? _redeemToBacking(msg.sender, principals, yields, recipient)
+                : _redeemToYieldBearing(msg.sender, principals, yields, recipient);
```
RedFox20, line 123: I don't know why linter formats it like this, we might as well just use if / else in that case? What do you think?

```suggestion
        if (isBackingToken) {
            return _redeemToBacking(msg.sender, principals, yields, recipient);
        } else {
            return _redeemToYieldBearing(msg.sender, principals, yields, recipient);
        }
```

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    function _depositYieldBearing(uint256 yieldTokenAmount) private returns (uint256) {
+        if (yieldTokenAmount == 0) {
+            revert ZeroYieldTokenAmount();
+        }
+
+        IERC20 yieldBearingToken = IERC20(tempusPool.yieldBearingToken());
```
RedFox20, line 131: Also cast no longer needed

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+
+        // Transfer funds from msg.sender to tempusPool
+        uint256 transferredYBT = yieldBearingToken.untrustedTransferFrom(
```
RedFox20, line 134: Since we are using `IERC20Metadata` now for `tempusPool.yieldBearingToken`, then you might have to add:
```
using UntrustedERC20 for IERC20Metadata;
```
But only if you start getting some errors. Not sure without changing things...

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+
+        return mintedShares;
+    }
+
+    function _depositBacking(uint256 _backingTokenAmount) private returns (uint256) {
+        if (_backingTokenAmount == 0) {
+            revert ZeroBackingTokenAmount();
+        }
+
+        IERC20 backingToken = IERC20(tempusPool.backingToken());
```
RedFox20, line 155: Here as well, the cast is no longer needed. However, it might need tobe `IERC20Metadata backingToken = ...`

test/otc/TempusOTC.test.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  async function acceptOfferDefault() {
+    await tempusOTC.acceptOffer(testPool, acceptOfferUser, acceptOfferSendAmount, false);
+  }
+
+  async function withdrawYieldAfterOfferAcceptedDefault() {
+    await tempusOTC.withdrawYieldAfterOfferAccepted(testPool, setOfferUser, yieldRequestedAmount);
+  }
+
+  async function setOfferChecks() {
```
RedFox20, line 46: I think these should need a comment which describes what `setOfferChecks` does, in brief
It's not clear to me at first glance, so a short comment could help readers with that 👍🏻

test/otc/TempusOTC.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    
+    expect(await tempusOTC.principalSetOfferAmount()).to.equal(setOfferAmount);
+    expect(await tempusOTC.yieldRequestedAmount()).to.equal(yieldRequestedAmount);
+    expect(await tempusOTC.offerStatus()).to.equal(OfferStatus.Created);
+  }
+
+  async function cancelOfferChecks() {
+    expect(await pool.yieldShare.balanceOf(tempusOTC.address)).to.equal(0);
+    expect(await pool.yieldShare.balanceOf(ownerUser)).to.equal(0);
+    expect(await pool.yieldShare.balanceOf(setOfferUser)).to.equal(setOfferAmount);
```
RedFox20, line 63: I see this triple user pattern a lot. How about creating a helper:

```ts
interface BalancesExpectation {
  tempusOTC: number;
  ownerUser: number;
  setOfferUser: number;
}

async function checkWallets(share: PoolShare, expected: BalancesExpectation ) {
    expect(await share.balanceOf(tempusOTC.address)).to.equal(expected.tempusOTC, "tempusOTC balance check failed");
    expect(await share.balanceOf(ownerUser)).to.equal(expected.ownerUser, "ownerUser balance check failed");
    expect(await share.balanceOf(setOfferUser)).to.equal(expected.setOfferUser, "setOfferUser balance check failed");
}
```

Usage:
```ts
await checkWallets(pool.yieldShare, { tempusOTC: 0, ownerUser: 0, setOfferUser: setOfferAmount });
```
Should save a bit of redundancy

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+        if (offerStatus != OfferStatus.NotSet) {
+            revert OfferAlreadySet();
+        }
+
+        uint256 mintedShares = isBackingToken ? _depositBacking(tokenAmount) : _depositYieldBearing(tokenAmount);
+
+        _fillOfferDetails(recipient, OfferStatus.Created, mintedShares, requestYieldAmount);
+
+        IERC20 yieldShares = IERC20(address(tempusPool.yieldShare()));
+        yieldShares.transfer(recipient, mintedShares);
```
mijovic, line 43: On latest master you can do this 
```suggestion
        tempusPool.yieldShare().transfer(recipient, mintedShares);
```

contracts/otc/TempusOTC.sol
mijovic: Here we transfer only `IPoolShare` I would skip adding reentrancy checks here

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+    uint256 public principalSetOfferAmount;
+    uint256 public yieldRequestedAmount;
+
+    constructor(ITempusPool pool) {
+        tempusPool = pool;
+        offerStatus = OfferStatus.NotSet;
+    }
+
+    /// @inheritdoc ITempusOTC
+    function setOffer(
```
mijovic, line 28: what if we make it private and just call from constructor directly (so the constructor itself becomes offer setter)?We can save one step here. Then offer details are immutable.

contracts/otc/TempusOTC.sol
```diff
@@ -0,0 +1,272 @@
...
+        if (offerStatus != OfferStatus.Created) {
+            revert OfferNotCreated();
+        }
+
+        uint256 principalLockedAmount = principalSetOfferAmount;
+
+        _fillOfferDetails(address(0), OfferStatus.NotSet, 0, 0);
+
+        IERC20 principalShares = IERC20(address(tempusPool.principalShare()));
+        principalShares.transfer(recipient, principalLockedAmount);
```
mijovic, line 59:
```suggestion
        tempusPool.principalShare().transfer(recipient, principalLockedAmount);
```

contracts/otc/TempusOTC.sol
mijovic: In this case canceling offer can sent tokens back and do self destruct

contracts/otc/ITempusOTC.sol
StefanVitor: Yes, I used Grammarly, but it didn't find any error about this :)

contracts/otc/TempusOTC.sol
StefanVitor: Agree, I'll change to if/else.

PR #594, opened 2022-04-08
Introduce custom errors to TempusPool contracts

pashov:
part of #190

### DIFF
--- a/contracts/ITempusController.sol
+++ b/contracts/ITempusController.sol
@@ -102,7 +102,7 @@ interface ITempusController {
 
     /// @dev Error thrown when the pool has already matured
     /// @param tempusPool The address of the pool that has already matured
-    error PoolAlreadyFinalized(address tempusPool);
+    error PoolAlreadyMatured(ITempusPool tempusPool);
 
     /// @dev Error thrown when trying to exit AMM before maturity but pricipal and yield token amounts are not equal
     /// @param principalTokenAmount The amount of principal tokens
@@ -113,13 +113,13 @@ interface ITempusController {
     /// @param sender The transfer sender
     /// @param recipient The transfer recipient
     /// @param amount The principal tokens amount
-    error FailedPrincipalTokensTransfer(address sender, address recipient, uint256 amount);
+    error FailedPrincipalTokensTransfer(address sender, ITempusController recipient, uint256 amount);
 
     /// @dev Error thrown when a yield tokens transfer fails
     /// @param sender The transfer sender
     /// @param recipient The transfer recipient
     /// @param amount The yield tokens amount
-    error FailedYieldTokensTransfer(address sender, address recipient, uint256 amount);
+    error FailedYieldTokensTransfer(address sender, ITempusController recipient, uint256 amount);
 
     /// @dev Error thrown when the yields rate is zero
     error ZeroYieldsRate();
--- a/contracts/ITempusPool.sol
+++ b/contracts/ITempusPool.sol
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.10;
 
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
 import "./token/IPoolShare.sol";
 import "./utils/IOwnable.sol";
 import "./utils/IVersioned.sol";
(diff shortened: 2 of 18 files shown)


pashov:
Currently blocked and waiting for amm v2 to be merged to master

contracts/ITempusPool.sol
```diff
@@ -44,6 +44,110 @@ interface ITempusFees is IOwnable {
...
+    /// @param underlyingToken The address of the underlying token
+    error MoreThan18DecimalsOfUnderlyingToken(address underlyingToken);
+
+    /// @dev Error thrown when the given token is not a Compound token
+    /// @param token The address of the given token
+    error NotACToken(address token);
+
+    /// @dev Error thrown when the Compound token does not have 8 decimals
+    /// @param cToken The address of the Compound token
+    error CTokenNotHaving8Decimals(address cToken);
```
axic, line 115: I think we could generalise these errors a bit, for example `InvalidBackingToken`, `InvalidBackingTokenDecimals(expected, actual)`, etc.

axic: Otherwise I wonder if the protocol-specific errors should just go directly into the pool implementation, as that is where they are only emitted.

contracts/ITempusController.sol
```diff
@@ -99,7 +99,7 @@ interface ITempusController {
 
     /// @dev Error thrown when the pool has already matured
     /// @param tempusPool The address of the pool that has already matured
-    error PoolAlreadyFinalized(address tempusPool);
+    error PoolAlreadyMatured(address tempusPool);
```
RedFox20, line 102: Is it possible to use `ITempusPool tempusPool` here instead, so we could avoid `address(tempusPool)` and `address(this)` casts?

RedFox20:
This is looking pretty good. 👍🏻 How far do you think you are from completion? I think it should be fine to only refactor one part of the codebase, review & merge the PR and then continue with other parts 🙂

contracts/ITempusPool.sol
pashov: I thought about this. My idea was that this way it is a lot easier to troubleshoot as you just need to look at the name of the error and you already know the problem. Of course this comes at the cost of lots more errors, which increases deployment gas costs and bytecode size.
Can we get some more opinions on this please, I'd like to figure out the best possible way to do this.
@RedFox20 @mijovic @yuval-weiss

contracts/ITempusController.sol
pashov: Nice, done

contracts/ITempusPool.sol
RedFox20: We have multiple errors like this, so generalizing makes sense. However, my question is how are the errors reported via ethers.js ? Unit tests must ensure the error message is exact (eg: `InvalidBackingTokenDecimals(expected:8, actual:6)`). If that's unfeasible, then we should stick to `CTokenMustHave8Decimals` etc

contracts/ITempusPool.sol
```diff
@@ -43,6 +43,110 @@ interface ITempusFees is IOwnable {
...
+    /// @param yieldTokenAmount The amount of yield tokens
+    error NotEqualPrincipalAndYieldTokenAmounts(uint256 principalTokenAmount, uint256 yieldTokenAmount);
+
+    /// @dev Error thrown when the underlying token has more than 18 decimals
+    /// @param underlyingToken The address of the underlying token
+    error MoreThan18DecimalsOfUnderlyingToken(address underlyingToken);
+
+    /// @dev Error thrown when the given token is not a Compound token
+    /// @param token The address of the given token
+    error NotACToken(address token);
```
StefanVitor, line 110: I would set this as ```error NotACToken(IERC20 token);```, so when we call this, we will not use address cast.

contracts/ITempusPool.sol
```diff
@@ -43,6 +43,110 @@ interface ITempusFees is IOwnable {
...
+
+    /// @dev Error thrown when the Compound token does not have 8 decimals
+    /// @param cToken The address of the Compound token
+    error CTokenNotHaving8Decimals(address cToken);
```
StefanVitor, line 114: Same here as it was "I would set this as error NotACToken(IERC20 token);, so when we call this, we will not use address cast."

contracts/ITempusPool.sol
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.10;
 
+import "./protocols/compound/ICErc20.sol";
```
StefanVitor, line 4: Maybe would be better that you import `import "@openzeppelin/contracts/token/ERC20/IERC20.sol";`, and after that, for token address, use IERC20.

contracts/ITempusPool.sol
yuval-weiss: Yeah I also think we should generalize some of these errors + move protocol-specific errors (that cannot be generalized) out of `ITempusPool.sol`.

pashov: Ok I removed lots of errors, added more abstract ones, and moved protocol-specific errors to their contracts.

contracts/ITempusController.sol
```diff
@@ -113,13 +113,13 @@ interface ITempusController {
     /// @param sender The transfer sender
     /// @param recipient The transfer recipient
     /// @param amount The principal tokens amount
-    error FailedPrincipalTokensTransfer(address sender, address recipient, uint256 amount);
+    error FailedPrincipalTokensTransfer(address sender, ITempusController recipient, uint256 amount);
```
mijovic, line 116: why `recipient` is `ITempusController`?

contracts/ITempusPool.sol
```diff
@@ -43,6 +43,80 @@ interface ITempusFees is IOwnable {
...
+
+    /// @dev Error thrown when the early redeem fee percentage is bigger than the maximum
+    /// @param earlyRedeemFeePercent The early redeem fee percent given
+    /// @param maximumEarlyRedeemFeePercent The maximum early redeem fee percent allowed
+    error EarlyRedeemFeePercentageTooBig(uint256 earlyRedeemFeePercent, uint256 maximumEarlyRedeemFeePercent);
+
+    /// @dev Error thrown when the mature redeem fee percentage is bigger than the maximum
+    /// @param matureRedeemFeePercent The mature redeem fee percentage given
+    /// @param maximumMatureRedeemFeePercent The maximum mature redeem fee percentage allowed
+    error MatureRedeemFeePercentageTooBig(uint256 matureRedeemFeePercent, uint256 maximumMatureRedeemFeePercent);
```
mijovic, line 80: Why not merge these into `FeePercentageTooBig` and maybe have another param to distinguish these 3?
It is basically the same error?

contracts/ITempusController.sol
RedFox20: We discussed this previously, to avoid casts `address(this)` and `address(controller)` we can just use the interface

contracts/pools/CompoundTempusPool.sol
```diff
@@ -17,6 +17,20 @@ contract CompoundTempusPool is TempusPool {
...
     bytes32 public constant override protocolName = "Compound";
 
+    /// @dev Error thrown when the call to the comptroller enter markets method fails
+    /// @param marketToken The address of the market token
+    error ComptrollerEnterMarketsFailed(ICErc20 marketToken);
+
+    /// @dev Error thrown when the minting of Compound tokens fails
+    /// @param cToken The address of the Compound token to mint
+    /// @param amountToMint The amount of Compound token to mint
+    error CTokenMintFailed(address cToken, uint256 amountToMint);
```
RedFox20, line 27: You could use `ICToken` here

contracts/ITempusController.sol
pashov: The old error message was `Principals transfer failed`
tbh I am not perfectly sure why in the code we send the principal tokens to the Controller smart contract, I guess it works like a burning mechanism?

contracts/pools/CompoundTempusPool.sol
pashov: I am not doing a cast where I am using the error. If I use `ICToken` I will have to cast there.

contracts/pools/CompoundTempusPool.sol
```diff
@@ -17,6 +17,20 @@ contract CompoundTempusPool is TempusPool {
 
     bytes32 public constant override protocolName = "Compound";
 
+    /// @dev Error thrown when the call to the comptroller enter markets method fails
+    /// @param marketToken The address of the market token
+    error ComptrollerEnterMarketsFailed(ICErc20 marketToken);
```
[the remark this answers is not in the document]
pashov: https://docs.soliditylang.org/en/v0.8.11/style-guide.html#order-of-layout
Solidity docs say it's a best practice to put events after state variables. I guess errors are very close to events that's why I used this approach. Of course we can choose a different convention

contracts/ITempusPool.sol
```diff
@@ -43,6 +43,71 @@ interface ITempusFees is IOwnable {
...
+
+    /// @dev Error thrown when the expected decimals of a token are more than the maximum expected decimals
+    /// @param token The address of the token
+    /// @param maximumExpectedDecimals The maximum expected decimals
+    /// @param actualDecimals The actual decimals
+    error MoreThanMaximumExpectedDecimals(address token, uint256 maximumExpectedDecimals, uint256 actualDecimals);
+
+    /// @dev Error thrown when the given token is not a valid one in the pool's context
+    /// @param token The address of the given token
+    error InvalidBackingToken(address token);
```
StefanVitor, line 103: Because using cast, put `error InvalidBackingToken(IERC20 token);`

contracts/pools/CompoundTempusPool.sol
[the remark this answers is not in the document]
mijovic: This is interesting. I would also lean towards listing them before constants/storage/functions, which I would do also for events. But, since docs are suggesting to do like this probably we leave for now.

contracts/ITempusController.sol
mijovic: Let's leave as is in that case, we can then find a better name. It is also not a scope of this PR tbh.

mijovic (approved):

PR #596, opened 2022-04-11
amm: Clear up amplification data helpers

axic:


contracts/amm/TempusAMM.sol
```diff
@@ -471,6 +471,12 @@ contract TempusAMM is ITempusAMM, ERC20, Pausable, Ownable {
 
     // Amplification
 
+    function setInitialAmplification(uint256 currentValue) private {
```
axic, line 474: This helper is needed if one wants to emit the two events in the constructor even in case of fixed amplification.

The reason one would want to do this, if they expect that amplification may be modified later.

axic:
@mijovic I think this approach takes less mental capacity to read/review, as the concerns are grouped (initial/start/stop which do events) and the low-level helpers.

contracts/amm/TempusAMM.sol
```diff
@@ -509,25 +519,25 @@ contract TempusAMM is ITempusAMM, ERC20, Pausable, Ownable {
             uint256 precision
         )
     {
-        (value, isUpdating) = _getAmplificationParameter();
+        (value, isUpdating) = getAmplificationParameter();
         precision = StableMath._AMP_PRECISION;
     }
 
-    function _getAmplificationParameter() private view returns (uint256 value, bool isUpdating) {
+    function getAmplificationParameter() private view returns (uint256 value, bool isUpdating) {
```
axic, line 526: Actually underscores cannot easily be removed, as this function clashes with the above.

contracts/amm/TempusAMM.sol
axic: Could also move the require statements for startValue range here.

contracts/amm/TempusAMM.sol
```diff
@@ -91,9 +91,9 @@ contract TempusAMM is ITempusAMM, ERC20, Pausable, Ownable {
 
         scalingFactor = Fixed256x18.ONE * 10**(18 - t0.decimals());
 
-        _setAmplificationData(amplificationStartValue);
-
-        if (amplificationStartValue != amplificationEndValue) {
+        if (amplificationStartValue == amplificationEndValue) {
```
axic, line 94: This change is actually wrong.

contracts/amm/TempusAMM.sol
```diff
@@ -91,9 +91,9 @@ contract TempusAMM is ITempusAMM, ERC20, Pausable, Ownable {
...
+            setInitialAmplification(amplificationStartValue);
+        } else {
             require(amplificationStartValue < amplificationEndValue, "invalid amp");
```
axic, line 97: This should be moved into `startAmplificationParameterUpdate` however as that function can be called outside of the constructor too.

contracts/amm/TempusAMM.sol
```diff
@@ -40,8 +40,8 @@ contract TempusAMM is ITempusAMM, ERC20, Pausable, Ownable {
     // rapidly: for example, by doubling the value every day it can increase by a factor of 8 over three days (2^3).
     uint256 private constant MIN_UPDATE_TIME = 1 days;
     uint256 private constant MAX_AMP_UPDATE_DAILY_RATE = 2;
-    uint256 private constant MIN_AMPLIFICATION = 1000;
-    uint256 private constant MAX_AMPLIFICATION = 5000000;
+    uint256 private constant MIN_AMPLIFICATION = StableMath._MIN_AMP * StableMath._AMP_PRECISION;
+    uint256 private constant MAX_AMPLIFICATION = StableMath._MAX_AMP * StableMath._AMP_PRECISION;
```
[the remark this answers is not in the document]
mijovic: I agree with this completely

contracts/amm/TempusAMM.sol
mijovic: This is a question of do we want to allow amp updates in a different direction. We use this in test for swaps actually (we move amp from 5 to 95 and then back to 5). Let's maybe remove this check at all?

contracts/TempusPool.sol
```diff
@@ -14,6 +14,7 @@ import "./math/Fixed256xVar.sol";
 import "./utils/Ownable.sol";
 import "./utils/UntrustedERC20.sol";
 import "./utils/Versioned.sol";
+import "./amm/TempusAMM.sol";
```
mijovic, line 17: Is this accidentally here?

contracts/amm/TempusAMM.sol
```diff
@@ -554,27 +564,26 @@ contract TempusAMM is ITempusAMM, ERC20, Pausable, Ownable {
...
         uint256 endValue,
         uint256 startTime,
         uint256 endTime
     ) private {
+        assert(
+            startValue <= type(uint64).max &&
+                endValue <= type(uint64).max &&
+                startTime <= type(uint64).max &&
+                endTime <= type(uint64).max
+        );
```
mijovic, line 578: Format is really weird, probably you just need to fix indentation of last 3 checks

axic: This was done by `yarn lint`.

RedFox20: Who ever wrote this opinionated linter should be fired, now we all suffer from it 😞

contracts/TempusPool.sol
axic: Yes, it was for something else I don't remember now.

mijovic (approved):

PR #600, opened 2022-04-12
pool: use ERC20 interfaces instead of address

RedFox20:


contracts/utils/FullERC20.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.10;
+
+import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+
+/// @dev Provides convenient wrappers to access IERC20Metadata functions
+//       from OZ's IERC20 addresses
+library FullERC20 {
```
axic, line 8: Hmm, perhaps `ERC20MetadataHelper` or something similar is more appropriate?

RedFox20: Too much of a mouthful 😄

contracts/TempusController.sol
```diff
@@ -314,10 +314,8 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, Versio
             revert ZeroYieldTokenAmount();
         }
 
-        IERC20 yieldBearingToken = IERC20(tempusPool.yieldBearingToken());
-
         // Transfer funds from msg.sender to tempusPool
-        uint256 transferredYBT = yieldBearingToken.untrustedTransferFrom(
+        uint256 transferredYBT = IERC20(tempusPool.yieldBearingToken()).untrustedTransferFrom(
```
mijovic, line 318: This is already `IERC20` I think there is no need to cast here
```suggestion
        uint256 transferredYBT tempusPool.yieldBearingToken().untrustedTransferFrom(
```

RedFox20: You are right, it's an unnecessary cast

contracts/utils/UntrustedERC20.sol
```diff
@@ -16,12 +16,12 @@ library UntrustedERC20 {
     /// @param value The requested amount.
     /// @return The actual amount of tokens transferred.
     function untrustedTransfer(
-        IERC20 token,
+        IERC20Metadata token,
```
yuval-weiss, line 19: no need to change this type from `IERC20` to `IERC20Metadata`. This library is generic and works for `IERC20` with no dependency on `IERC20Metadata`. Also, even if we keep the type as `IERC20`, passing an `IERC20Metadata` will work to as it will be implicitly casted to `IERC20`

yuval-weiss: (same applies for all the other methods in this file)

mijovic: Other than this, we could merge it

RedFox20: Thanks! I fixed it 👍🏻 💯

mijovic (dismissed):

mijovic:
You need to solve conflicts unfortunately

mijovic (approved):

contracts/pools/CompoundTempusPool.sol
```diff
@@ -85,7 +85,7 @@ contract CompoundTempusPool is TempusPool {
         uint256 ybtBefore = balanceOfYBT();
 
         // Deposit to Compound
-        IERC20(backingToken).safeIncreaseAllowance(yieldBearingToken, amountBT);
+        IERC20(backingToken).safeIncreaseAllowance(address(yieldBearingToken), amountBT);
```
axic: Left a cast here 😅

contracts/stats/Stats.sol
```diff
@@ -18,8 +18,8 @@ contract Stats is ChainlinkTokenPairPriceFeed, Versioned {
     /// @param tempusPool The TempusPool to fetch its TVL (total value locked)
     /// @return total value locked of a TempusPool (denominated in BackingTokens)
     function totalValueLockedInBackingTokens(ITempusPool tempusPool) public view returns (uint256) {
-        PoolShare principalShare = PoolShare(address(tempusPool.principalShare()));
-        PoolShare yieldShare = PoolShare(address(tempusPool.yieldShare()));
+        IPoolShare principalShare = tempusPool.principalShare();
```
axic, line 21: Oh, should have also changed the import to `IPoolShare`.

PR #601, opened 2022-04-13
Introduce custom errors to Tempus AMM

### LINKED ISSUE #190: Switch to error types/codes from error strings

pashov:
Fixes #190

### DIFF
--- a/contracts/amm/interfaces/ITempusAMM.sol
+++ b/contracts/amm/interfaces/ITempusAMM.sol
@@ -42,6 +42,90 @@ interface ITempusAMM is IERC20, IRateProvider, IOwnable {
     /// @param amountOut Amount of tokensOut received by user
     event Swap(IPoolShare tokenIn, uint256 amountIn, uint256 amountOut);
 
+    /// @dev Error thrown when proposed swap fee is bigger than the maximum
+    /// @param swapFee The swap fee
+    /// @param maxSwapFee The maximum swap fee
+    error SwapFeeTooBig(uint256 swapFee, uint256 maxSwapFee);
+
+    /// @dev Error thrown when the two pool share tokens have different number of decimals
+    /// @param token0 The first token
+    /// @param token1 The second token
+    error TokenDecimalsMismatch(IPoolShare token0, IPoolShare token1);
+
+    /// @dev Error thrown when starting amplification value is bigger than the ending amplification value
+    /// @param startingAmplificationValue The starting amplification value
+    /// @param endingAmplificationValue The ending amplification value
+    error StartingAmplificationValueBiggerThanEndingAmplificationValue(
+        uint256 startingAmplificationValue,
+        uint256 endingAmplificationValue
+    );
+
+    /// @dev Error thrown when amplification value is smaller than the minimum
+    /// @param amplificationValue The starting amplification value
+    /// @param minAmplificationValue The minimum amplification value
+    error AmplificationValueTooSmall(uint256 amplificationValue, uint256 minAmplificationValue);
+
+    /// @dev Error thrown when amplification value is bigger than the maximum
+    /// @param amplificationValue The starting amplification value
+    /// @param maxAmplificationValue The maximum amplification value
+    error AmplificationValueTooBig(uint256 amplificationValue, uint256 maxAmplificationValue);
+
+    /// @dev Error thrown when the amount of a token is zero
+    error ZeroTokenAmount();
+
+    /// @dev Error thrown when LP tokens to receive are less than the minimum expected when adding liquidity
+    /// @param lpTokensOut The LP tokens to receive
+    /// @param minLPTokensOut The minimum LP tokens to receive
+    error AddingLiquidityLPTokensSlippage(uint256 lpTokensOut, uint256 minLPTokensOut);
+
+    /// @dev Error thrown when pool tokens to receive are less than the minimum expected when removing liquidity
+    /// @param poolTokensOut The pool tokens to receive
+    /// @param minPoolTokensOut The minimum pool tokens to receive
+    error RemovingLiquidityPoolTokensSlippage(uint256 poolTokensOut, uint256 minPoolTokensOut);
+
+    /// @dev Error thrown when the LP tokens to put in are more than the maximum expected when removing liquidity
+    /// @param lpTokensIn The LP tokens to put in
+    /// @param maxLpTokensIn The maximum LP tokens to put in
+    error RemovingLiquidityLpTokensSlippage(uint256 lpTokensIn, uint256 maxLpTokensIn);
+
+    /// @dev Error thrown when the tokens to receive are less than the minimum expected when doing a swap
+    /// @param tokensOut The tokens to receive
+    /// @param minTokensOut The minimum tokens to receive
+    error SwapGivenTokensInSlippage(uint256 tokensOut, uint256 minTokensOut);
+
+    /// @dev Error thrown when the tokens to put in are more than the maximum expected when doing a swap
+    /// @param tokensIn The tokens to put in
+    /// @param maxTokensIn The maximum tokens to put in
+    error SwapGivenTokensOutSlippage(uint256 tokensIn, uint256 maxTokensIn);
+
+    /// @dev Error thrown when the swap deadline has passed
+    /// @param deadline The swap deadline
+    /// @param currentTime The current timestamp
+    error SwapDeadlinePassed(uint256 deadline, uint256 currentTime);
+
+    /// @dev Error thrown when the token in is invalid
+    /// @param tokenIn The token in
+    error InvalidTokenIn(IPoolShare tokenIn);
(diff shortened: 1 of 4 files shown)


contracts/amm/interfaces/ITempusAMM.sol
```diff
@@ -42,6 +42,84 @@ interface ITempusAMM is IERC20, IRateProvider, IOwnable {
...
+    /// @param startingAmplificationValue The starting amplification value
+    /// @param minAmplificationValue The minimum amplification value
+    error StartingAmplificationValueTooSmall(uint256 startingAmplificationValue, uint256 minAmplificationValue);
+
+    /// @dev Error thrown when starting amplification value is bigger than the maximum
+    /// @param startingAmplificationValue The starting amplification value
+    /// @param maxAmplificationValue The maximum amplification value
+    error StartingAmplificationValueTooBig(uint256 startingAmplificationValue, uint256 maxAmplificationValue);
+
+    /// @dev Error thrown when swap fee is bigger than the maximum
```
mijovic, line 55:
```suggestion
    /// @dev Error thrown when proposed swap fee is bigger than the maximum
```

contracts/amm/interfaces/ITempusAMM.sol
```diff
@@ -42,6 +42,84 @@ interface ITempusAMM is IERC20, IRateProvider, IOwnable {
...
+    /// @param swapFee The swap fee
+    /// @param maxSwapFee The maximum swap fee
+    error SwapFeeTooBig(uint256 swapFee, uint256 maxSwapFee);
+
+    /// @dev Error thrown when the two given tokens have different number of decimals
+    /// @param token0Decimals The decimals of the first token
+    /// @param token1Decimals The decimals of the second token
+    error TokenDecimalsMismatch(uint256 token0Decimals, uint256 token1Decimals);
```
mijovic, line 63: I think it is better to set actual token addresses instead of decimals.
catching this error is useless probably. The other way you can easily query actual decimals and everything. 
wdyt?

contracts/amm/interfaces/ITempusAMM.sol
```diff
@@ -42,6 +42,84 @@ interface ITempusAMM is IERC20, IRateProvider, IOwnable {
...
+
+    /// @dev Error thrown when starting amplification value is bigger than the ending amplification value
+    /// @param startingAmplificationValue The starting amplification value
+    /// @param endingAmplificationValue The ending amplification value
+    error StartingAmplificationValueBiggerThanEndingAmplificationValue(
+        uint256 startingAmplificationValue,
+        uint256 endingAmplificationValue
+    );
```
mijovic, line 71: I was leaning towards deleting this error altogether along with removing revert. We can do this in a new PR

contracts/amm/interfaces/ITempusAMM.sol
```diff
@@ -42,6 +42,84 @@ interface ITempusAMM is IERC20, IRateProvider, IOwnable {
...
+
+    /// @dev Error thrown when ending amplification value is less than the minimum
+    /// @param endingAmplificationValue The ending amplification value
+    /// @param minAmplificationValue The minimum ending amplification value
+    error EndingAmplificationValueTooSmall(uint256 endingAmplificationValue, uint256 minAmplificationValue);
+
+    /// @dev Error thrown when ending amplification value is more than the maximum
+    /// @param endingAmplificationValue The ending amplification value
+    /// @param maxAmplificationValue The maximum ending amplification value
+    error EndingAmplificationValueTooBig(uint256 endingAmplificationValue, uint256 maxAmplificationValue);
```
mijovic, line 105: Maybe grouping with the same errors for `StartingAmplification`. Also, I vote for having only 2 types of errors here called `AmplificationValueTooSmall` and `AmplificationValueTooBig`

contracts/amm/interfaces/ITempusAMM.sol
```diff
@@ -42,6 +42,84 @@ interface ITempusAMM is IERC20, IRateProvider, IOwnable {
...
+
+    /// @dev Error thrown when amplification value update end time is too close to the current time
+    /// @param updateTimeRemaining The update time remaining
+    /// @param minUpdateTimeRemaining The minimum update time remaining
+    error AmplificationValueUpdateEndTimeTooClose(uint256 updateTimeRemaining, uint256 minUpdateTimeRemaining);
+
+    /// @dev Error thrown when the amplification value has an ongoing update
+    error AmplificationValueOngoingUpdate();
```
mijovic, line 113:
```suggestion
    error AmplificationOngoingUpdate();
```

contracts/amm/interfaces/ITempusAMM.sol
```diff
@@ -42,6 +42,84 @@ interface ITempusAMM is IERC20, IRateProvider, IOwnable {
...
+    error AlreadyInitialised();
+
+    /// @dev Error thrown when AMM is not initialised yet
+    error NotInitialisedYet();
+
+    /// @dev Error thrown when both of the given tokens have zero amount
+    error BothTokensWithZeroAmount();
+
+    /// @dev Error thrown when the price or LP slippage is more than the maximum tolerable
+    error Slippage();
```
mijovic, line 86: We will need to split this one into multiple error types I think. There is slippage for swaps, joins, and exits.

contracts/amm/interfaces/ITempusAMM.sol
pashov: Ok that makes sense. Do you think my mindset with errors should be "front-end handling first"?

contracts/amm/interfaces/ITempusAMM.sol
pashov: Added. Won't add it in all places because it clutters things too much. I try to shave off words that make the meaning of the docs the same even when they aren't there, but in some cases it just sounds weird so I fix it on a case-by-case basis. I will add the word here.

contracts/amm/interfaces/ITempusAMM.sol
pashov: Leaving this for another PR then.

contracts/amm/interfaces/ITempusAMM.sol
```diff
@@ -42,6 +42,84 @@ interface ITempusAMM is IERC20, IRateProvider, IOwnable {
...
+    );
+
+    /// @dev Error thrown when the amount of tokens is zero
+    error ZeroTokenAmount();
+
+    /// @dev Error thrown when AMM is already initialised
+    error AlreadyInitialised();
+
+    /// @dev Error thrown when AMM is not initialised yet
+    error NotInitialisedYet();
```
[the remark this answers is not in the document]
pashov: `NotInitialisedYet` is still used

contracts/amm/interfaces/ITempusAMM.sol
mijovic: frontend handling, or also someone using our contracts and having `try/catch` block, I think it is much more useful if someone catches error with actual addresses of tokens

contracts/amm/interfaces/ITempusAMM.sol
```diff
@@ -42,6 +42,86 @@ interface ITempusAMM is IERC20, IRateProvider, IOwnable {
...
+    /// @param minPoolTokensOut The minimum pool tokens to receive
+    error RemovingLiquidityPoolTokensSlippage(uint256 poolTokensOut, uint256 minPoolTokensOut);
+
+    /// @dev Error thrown when the LP tokens to put in are more than the maximum expected when removing liquidity
+    /// @param lpTokensIn The LP tokens to put in
+    /// @param maxLpTokensIn The maximum LP tokens to put in
+    error RemovingLiquidityLpTokensSlippage(uint256 lpTokensIn, uint256 maxLpTokensIn);
+
+    /// @dev Error thrown when the slippage when executing a swap is more than the maximum tolerable percentage
+    error SwapSlippage();
```
mijovic, line 95: what about?
```suggestion
    error SwapSlippage(uint256 requestedMinimum, uint256 amountReturned);
```

contracts/amm/interfaces/ITempusAMM.sol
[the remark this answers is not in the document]
mijovic: Ah, I see. We can try removing it in next PR

contracts/amm/interfaces/ITempusAMM.sol
mijovic: what about this?
```suggestion
    error SwapSlippage(SwapType swapType, uint256 slippageParam, uint256 amountSpentOrReceived);
```

pashov: Discussed in person.

mijovic (dismissed):

mijovic (approved):

PR #602, opened 2022-04-13
amm: Removing TempusAMM init method.

mijovic:
Since one can call either `join` or `exit` I was thinking why not merge them to one method, where init logic is done when no liquidity, and join logic otherwise

### DIFF
--- a/contracts/amm/TempusAMM.sol
+++ b/contracts/amm/TempusAMM.sol
@@ -97,38 +97,12 @@ contract TempusAMM is ITempusAMM, ERC20, Pausable, Ownable {
 
     // Adding liquidity
 
-    function init(uint256 amountToken0, uint256 amountToken1) external override {
-        require(amountToken0 != 0 && amountToken1 != 0, "token amounts can't be 0");
-        require(totalSupply() == 0, "Already initialised");
-
-        (uint256 amountIn0, uint256 amountIn1) = getRateAdjustedAmounts(
-            amountToken0,
-            amountToken1,
-            token0.getPricePerFullShare(),
-            token1.getPricePerFullShareStored()
-        );
-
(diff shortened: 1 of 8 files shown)


contracts/amm/TempusAMM.sol
```diff
@@ -97,54 +97,39 @@ contract TempusAMM is ITempusAMM, ERC20, Pausable, Ownable {
...
-            balance0,
-            balance1,
-            amountIn0,
-            amountIn1,
-            totalSupply(),
-            swapFeePercentage
-        );
+        uint256 lpTokensOut;
+        if (totalSupply() == 0) {
+            require(amountToken0 != 0 && amountToken1 != 0, "token amounts can't be 0");
```
mijovic, line 115: In the case of init, both tokens need to be non-zero.

RedFox20 (approved):

contracts/amm/interfaces/ITempusAMM.sol
```diff
@@ -48,11 +48,6 @@ interface ITempusAMM is IERC20, IRateProvider, IOwnable {
     /// second token in TempusAMM pair
     function token1() external view returns (IPoolShare);
 
-    /// Initialises TempusAMM by adding initial liquidity it
-    /// @param amountToken0 Amount of token0 to init TempusAMM with
-    /// @param amountToken1 Amount of token0 to init TempusAMM with
-    function init(uint256 amountToken0, uint256 amountToken1) external;
-
     /// Adds liquidity to TempusAMM
```
axic, line 51: Would have added here a note about the new special behaviour in case of `totalSupply == 0`.

mijovic: https://github.com/tempus-finance/tempus-protocol/pull/611

PR #604, opened 2022-04-14
link PositionManager to a specific TempusController

yuval-weiss:


contracts/PositionManager.sol
```diff
@@ -17,20 +17,28 @@ contract PositionManager is IPositionManager, ERC721, ReentrancyGuard {
...
 
     // solhint-disable-next-line no-empty-blocks
-    constructor(string memory name, string memory symbol) ERC721(name, symbol) {}
+    constructor(
+        address tempusController,
+        string memory name,
+        string memory symbol
+    ) ERC721(name, symbol) {
+        if (tempusController == address(0)) {
+            revert InvalidTempusController();
```
mijovic, line 31: Can you add test that fails here?

yuval-weiss: done

mijovic (approved):

axic:
@mijovic @yuval-weiss what is the benefit this provides? At the time of calling the pool it should still check against the accepted controller list.

PR #606, opened 2022-04-14
new `Decimal` util with fixed precision

RedFox20:
By using fixed precision, this meshes very well with Solidity contracts through ethers.

The reason why I decided to stop using `decimal.js` was because
a) it has dynamic precision, which means fractional precision can increase and thus give different results from Solidity math
b) there are no convenient adapters for add/sub/mul/div and BigNumber

Writing our own FixedPoint decimal was quite easy (1 day of work), and provides ability to perform FixedPoint math before passing the value to a Contract, which is now directly supported by existing utilities.

The underlying implementation uses JS `bigint` and doesn't rely on anything from ethers BigNumber. This will theoretically make it easier to target something other than Solidity contracts through ethers.js


axic:
So this should close #258, right?

RedFox20:
> So this should close #258, right?

Yeah, I would say this Closes #258, although the `toBigNum` would remain. Maybe we can give it a better name instead? `toContractDecimal` ?

test/utils/Decimal.ts
```diff
@@ -1,7 +1,109 @@
 import { ethers } from "hardhat";
 import { BigNumber } from "ethers";
 
-export type Numberish = Number | string;
+
+/**
+ * Matches the most common ERC20 18-decimals precision, such as ETH
```
axic, line 7: Strictly speaking Eth is not ERC-20 😅

test/utils/Decimal.ts
```diff
@@ -1,7 +1,109 @@
...
+  */
+export function decimal(value:Numberish, maxDecimals:number = DEFAULT_DECIMAL_PRECISION): Decimal {
+  return new Decimal(value, maxDecimals);
+}
+ 
+/**
+ * @abstract A Fixed-Point Decimal type with a strongly defined `decimals` precision
+ *           compatible with ERC20.decimals() concept.
+ */
+export class Decimal {
```
axic, line 25: Should we call this FixedPoint to be closer to the Solidity counterparts?

test/utils/Decimal.ts
```diff
@@ -1,7 +1,109 @@
...
+  bn: BigNumber; // big integer that holds the FixedPoint Decimal value
+  decimals: number; // number of decimal digits that form a fraction, can be 0
+
+  constructor(value:Numberish, decimals:number) {
+    this.decimals = decimals;
+    this.bn = Decimal.toFixedPointInteger(value, decimals);
+  }
+
+  public static toFixedPointInteger(value:Numberish, decimals:number): BigNumber {
+    // accept BigNumber without any validation, assume Dev knows what they're doing
```
axic: Bold statement.

test/utils/Decimal.ts
```diff
@@ -1,7 +1,109 @@
...
+export class Decimal {
+  bn: BigNumber; // big integer that holds the FixedPoint Decimal value
+  decimals: number; // number of decimal digits that form a fraction, can be 0
+
+  constructor(value:Numberish, decimals:number) {
+    this.decimals = decimals;
+    this.bn = Decimal.toFixedPointInteger(value, decimals);
+  }
+
+  public static toFixedPointInteger(value:Numberish, decimals:number): BigNumber {
```
axic: Do we need this to be public or would the other `decimal` hrlper be enough?

test/utils/Decimal.ts
mijovic: I agree. Decimal suggests maybe floating point?

test/utils/Decimal.ts
RedFox20: wETH *

test/utils/Decimal.ts
RedFox20: If you pass a BigNumber to this, it means you want manual control. And it's actually necessary for some of the interop to work :)

test/utils/Decimal.ts
RedFox20: That was changed in a newer commit, it is now private

test/utils/Decimal.ts
RedFox20: Right now it syncs well with the `decimal()` utility function, so creating these is really convenient:
```ts
let x = decimal(209.99).mul(2).sub(20);
```

RedFox20: So as an alternative probably this would work similarly:
```ts
let x: Fixed = fixed(209.99).mul(2).sub(20);
```
Or alternatively if we keep the function name as `fixed()`, then FP will work as well:
```ts
let x: FixedPoint = fixed(209.99).mul(2).sub(20);
```

mijovic: I don't have preference when you list like this all options look good to me :)

test/utils/Decimal.ts
```diff
@@ -1,7 +1,201 @@
...
+  }
+
+  /** @return decimal(this) / decimal(x) */
+  public div(x:Numberish): Decimal {
+    // divf = (a * ONE) / b
+    return this.toDecimal( (this.int * this.one()) / this.toDecimalBN(x) );
+  }
+
+  private static toFixedPointInteger(value:Numberish, decimals:number): bigint {
+    if (typeof(value) === "bigint") {
```
axic, line 118: `instanceof` doesn't work for this as `bigint` as a type?

test/utils/Decimal.ts
```diff
@@ -1,7 +1,201 @@
...
+  public toRounded(maxDecimals:number): string {
+    return this._toString(maxDecimals, true);
+  }
+
+  public toJSON(key?: string): any {
+      return { type: "Decimal", value: this.toString() };
+  }
+
+  public toNumber(): number {
+    return Number(this.toString());
```
axic, line 73: Should this throw in case of truncation?

test/utils/Decimal.ts
```diff
@@ -1,7 +1,201 @@
...
+  }
+
+  private static readonly ONE_CACHE: { [key:number]: bigint } = {
+    6: BigInt("1000000"),
```
axic, line 77: Doesn't the `1000000n` work as a constructor for `bigint`?

test/utils/Decimal.ts
```diff
@@ -1,7 +1,201 @@
...
+    Object.freeze(this);
+  }
+
+  /** @returns BigNumber from this Decimal */
+  public toBigNumber(): BigNumber {
+    return BigNumber.from(this.int.toString());
+  }
+
+  private toDecimalBN(x:Numberish): bigint {
+    return Decimal.toFixedPointInteger(x, this.decimals);
```
axic, line 41: To be honest I have been confused by this reading the code, perhaps `Decimal.fromNumberish` is more fitting?

test/utils/Decimal.ts
```diff
@@ -1,7 +1,201 @@
...
+ *           compatible with ERC20.decimals() concept.
+ */
+export class Decimal {
+  readonly int: bigint; // big integer that holds the FixedPoint Decimal value
+  readonly decimals: number; // number of decimal digits that form a fraction, can be 0
+
+  constructor(value:Numberish, decimals:number) {
+    this.decimals = decimals;
+    this.int = Decimal.toFixedPointInteger(value, decimals);
+    Object.freeze(this);
```
axic, line 32: Noice!

test/utils/Decimal.ts
```diff
@@ -1,7 +1,201 @@
...
+  }
+
+  /** @returns BigNumber from this Decimal */
+  public toBigNumber(): BigNumber {
+    return BigNumber.from(this.int.toString());
```
axic, line 37: According to ethers.js docs this support bigint as an input, i.e. 
```suggestion
    return BigNumber.from(this.int);
```

Did you try that?

test/utils/Decimal.ts
```diff
@@ -1,7 +1,201 @@
...
+  }
+
+  private toDecimalBN(x:Numberish): bigint {
```
axic, line 40: `intoBigint` ?

test/utils/Decimal.ts
```diff
@@ -1,7 +1,201 @@
...
+    return Number(this.toString());
+  }
+
+  private static readonly ONE_CACHE: { [key:number]: bigint } = {
+    6: BigInt("1000000"),
+    18: BigInt("1000000000000000000"),
+  };
+
+  /** 1.0 expressed as a scaled bigint */
+  public one(): bigint {
```
axic, line 82: Does this needs to be public given it is `bigint`?

test/utils/Decimal.ts
```diff
@@ -1,7 +1,201 @@
...
+
+    if (decimalIdx === -1) { // input was integer eg "1234"
+      return BigInt(val.padEnd(val.length + decimals, "0"));
+    }
+
+    // input was a decimal eg "123.45678"
+    // extract the fractional part of the decimal string
+    const fract = val.slice(decimalIdx+1, Math.min(decimalIdx+1+decimals, val.length));
+    // if it's not long enough, create a trail of zeroes to satisfy decimals precision
+    const trail = decimals > fract.length ? decimals - fract.length : 0;
```
axic, line 159: This needs to address the case where `fract.length > decimals`.

test/utils/Decimal.ts
```diff
@@ -1,7 +1,201 @@
...
+
+  private _toString(maxDecimals:number, round:boolean=false): string {
+    if (this.decimals === 0) {
+      return this.int.toString();
+    }
+
+    // get the BN digits and check if it's negative
+    const s = this.int.toString();
+    const neg = s[0] === '-';
+    const abs = neg ? s.slice(1) : s;
```
axic, line 171: Sup? Why is all these done on strings?

```suggestion
    const neg = this.int < 0n;
    const abs = neg ? (this.int * -1n) : this.int;
```

test/Decimal.test.ts
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    // easier to test init and toString together
+    it("init/toString", () =>
+    {
+      equals(dec18('100.123456789012345678'), '100.123456789012345678');
+      equals(dec6(10.0), '10.000000');
+      equals(dec6(Number(10.0)), '10.000000');
+      equals(dec6(123.4567890), '123.456789');
+      equals(dec6('123.4567890'), '123.456789');
```
axic, line 29: Should have a test for `dec6('123.45678901')`

test/utils/Decimal.ts
RedFox20: Probably not. Losing precision should be expected when moving from Decimal to a number. For more controlled use cases we have `toTruncated(): string`.

I added documentation to highlight that truncation will happen. But I don't think crashing here is a good solution

test/utils/Decimal.ts
RedFox20: I got an error when I originally tried to do it:
`BigInt literals are not available when targeting lower than ES2020.ts(2737)`

axic: Strange, but it is not that important, either should work.

test/utils/Decimal.ts
RedFox20: `Decimal.fromNumberish`, the `from` would imply we are creating a `Decimal`, but this is creating a `bigint`.

So `toFixedPointInteger` seems like a very accurate name.

`toDecimalBN` on the other hand is confusing. It's a wrapper with a bit of a bad name. Should definitely rename it to something else like `toFixedPoint` ?

test/utils/Decimal.ts
RedFox20: Immutable structs 🙂

test/utils/Decimal.ts
RedFox20: Yeah, indeed I tried that. I checked their source code and the reason I chose `toString()` there was to just make it faster.
Their implementation for `bigint` support is recursive:
```ts
        if (typeof(anyValue) === "bigint") {
            return BigNumber.from(anyValue.toString());
        }
```
So passing `string` is actually the most straightforward way. Everything is kept as a string internally for `BigNumber`

axic: JavaScript 😢

test/utils/Decimal.ts
RedFox20: Yeah that might be better. But another thing which is important here is that the bigint has the correct fixed point precision.

So something like `toFixedPoint()` would be more accurate? WDYT?

test/utils/Decimal.ts
RedFox20: I decided to leave this public just in case someone wants to do their own bigint math manually. This would provide the necessary `ONE` constant for that. But maybe that is unnecessary and I should make this private?

test/utils/Decimal.ts
RedFox20: I tried with `toFixedPointInt()`

test/utils/Decimal.ts
RedFox20: It's not very obvious but it is handled by the slice 1 line above. I'll add an extra comment. This is also covered by unit tests 👍🏻 


The trick is from the slice end: `slice(start?: number, end?: number): string;`

```ts
end: Math.min(decimalIdx+1+decimals, val.length)
```
It slices up to `decimals` characters of the fraction, including the decimal point character `+1`.
Or if the string is shorter, it terminates at original string's end

test/utils/Decimal.ts
RedFox20: Mostly because the rest of the method operates on the `abs: string`. It would have to look like this:
```ts
    const neg = this.int < 0;
    const abs = (neg ? this.int * BigInt(-1) : this.int).toString();
```
Probably this.int * BigInt(-1) will be faster because s.slice(1) makes a new substring

test/utils/Decimal.ts
```diff
@@ -37,8 +44,9 @@ export class Decimal {
     return BigNumber.from(this.int.toString());
   }
 
-  private toDecimalBN(x:Numberish): bigint {
-    return Decimal.toFixedPointInteger(x, this.decimals);
+  /** @returns Numberish converted to this Decimal precision bigint */
+  private toFixedPointInt(x:Numberish): bigint {
```
axic, line 48: Hmm, new name idea: `toScaledBigInt`? But yeah it doesn't matter.

test/utils/Decimal.ts
```diff
@@ -50,9 +268,16 @@ export function formatDecimal(bigDecimal:BigNumber, decimalBase:number): Numberi
   return str;
 }
 
+/**
+ * Truncates a number directly into a BigNumber, without any scaling
+ */
+export function bn(number:Numberish): BigNumber {
```
axic, line 274: Is this used for anything?

test/utils/Decimal.ts
RedFox20: Oh, that's a good one 💯 👍🏻 I'll use that

test/utils/Decimal.ts
RedFox20: Actually no, it's not used by anything, I just added it for potential future use. Idea was that you won't have to import `BigNumber` type in your typescript files and just use this utility:
```ts
this.contract.startAmplification( bn(1000) );
```
WDYT?

axic: I think it is not really related to this library. Can't wait for the refactoring dropping all the legacy decimal helpers though. Perhaps do it as an independent PR and we discuss the uses there?

RedFox20: Yeah it can be easily removed and added in later

axic (approved):

Issue #470, opened 2021-12-11
Add ERC20Permit (EIP-2612) support to PoolShares


axic:
And at the same time support them to be passed in the controller. This would mean only `deposit` would need an external approval for the BT/YBT, everything else could be handled in-system. The permit itself is done via signature verification which isn't entirely cheap, likely the cost here would be pretty similar, but it can be a better user experience.

PR #614, opened 2022-04-25
ERC20Permit support

mijovic:
Fixes https://github.com/tempus-finance/tempus-protocol/issues/470


StefanVitor (approved):
In my opinion, all is good.

axic:
I would say this is not the end of it, because the controller must also suppirt passing through permits in order to make this useful.

PR #620, opened 2022-05-14
PositionManager: ensure that deposit with Ether matches parameters

axic:


contracts/PositionManager.sol
```diff
@@ -45,7 +45,9 @@ contract PositionManager is IPositionManager, ERC721, ReentrancyGuard {
                 ? tempusPool.backingToken()
                 : tempusPool.yieldBearingToken();
 
-            if (address(depositedAsset) != address(0)) {
+            if (address(depositedAsset) == address(0)) {
+                assert(tokenAmountToDeposit == params.tokenAmountToDeposit);
```
yuval-weiss, line 49: I think reverting with an error is more suitable here

yuval-weiss (dismissed):

contracts/PositionManager.sol
```diff
@@ -45,7 +45,11 @@ contract PositionManager is IPositionManager, ERC721, ReentrancyGuard {
                 ? tempusPool.backingToken()
                 : tempusPool.yieldBearingToken();
 
-            if (address(depositedAsset) != address(0)) {
+            if (address(depositedAsset) == address(0)) {
+                if (tokenAmountToDeposit != params.tokenAmountToDeposit) {
+                    revert EtherDepositMismatch();
+                }
+            } else {
                 tokenAmountToDeposit = depositedAsset.untrustedTransferFrom(
```
axic, line 53: And it this case it should check that `msg.value == 0`.

axic: Ah I remember, you pass the value nonetheless, so that delegates the check properly to the pool.

So then we just need tests for this.

RedFox20: @axic ?

axic: Yes please, sir.

PR #621, opened 2022-05-14
Test type improvements

axic:


test/utils/TempusPool.ts
```diff
@@ -416,7 +416,7 @@ export class TempusPool extends ContractBase {
   /**
    * @returns True if maturity has been reached and the pool was finalized.
    */
-  async matured() {
+  async matured(): Promise<Boolean> {
     return this.contract.matured();
```
axic, line 420: @RedFox20 what is the deciding factor between `return` and `return await`? We use both.

RedFox20: In this case, my opinion now is that we should always set the return type as a `Promise<T>` and do a return instead of awaiting. Especially when there is only one Promise<> being returned like in this example. The reasoning here is that JS/TS is not able to optimize away the extra await.

So in the case of just returning an existing promise, ideally the function does less work.

RedFox20:
This looks great 💯

RedFox20 (approved):

PR #617, opened 2022-04-27
Replace Versioned interface with ERC165 support

pashov:
Fixes #240

PR #630, opened 2022-05-19
Fix ERC165 standart compatibility

pashov:
https://github.com/tempus-finance/tempus-protocol/pull/617/files#r864567523

### DIFF
--- a/contracts/ITempusPool.sol
+++ b/contracts/ITempusPool.sol
@@ -2,7 +2,6 @@
 pragma solidity 0.8.10;
 
 import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
-import "@openzeppelin/contracts/utils/introspection/IERC165.sol";
 
 import "./token/IPoolShare.sol";
 import "./utils/IOwnable.sol";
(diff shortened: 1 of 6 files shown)


contracts/ITempusPool.sol
```diff
@@ -2,7 +2,6 @@
 pragma solidity 0.8.10;
 
 import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
-import "@openzeppelin/contracts/utils/introspection/IERC165.sol";
```
axic, line 5: What is importing this now so that it can be removed?

pashov: TempusPool.sol at line 8 and 25. Do you think it should stay in the interface

contracts/ITempusPool.sol
```diff
@@ -44,7 +43,7 @@ interface ITempusFees is IOwnable {
 }
 
 /// All state changing operations are restricted to the controller.
-interface ITempusPool is ITempusFees, IERC165 {
```
mijovic, line 47: why did we remove it here?

pashov: Yes. This is because the standard says "inherit ERC165" but when I do inherit it in the implementation and if I have the interface inherited in the interface then there is a collision of methods

mijovic (approved):

PR #633, opened 2022-05-27
refactor: set up `tempus-sdk` directory

RedFox20:
This is how we last discussed the SDK should be set up.

What is missing here is NPM package generation that frontend guys could consume. I added frontend team to review so they can get a picture of what to expect.


mijovic:
Looks good to me, we just need to fix the tests

mijovic (approved):

PR #636, opened 2022-05-29
Add permit support to controller

mijovic:


contracts/TempusController.sol
```diff
@@ -178,12 +178,25 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, ERC165
...
         uint256 principalsStaked,
         uint256 yieldsStaked,
+        ERC20PermitSignature calldata lpPermit,
         uint256 maxLpTokensToRedeem,
         bool toBackingToken
     ) external override nonReentrant returns (uint256) {
         requireRegistered(address(tempusAMM));
         requireRegistered(address(tempusPool));
 
+        if (lpPermit.owner != address(0)) {
```
mijovic, line 188: Is it ok like this? Do we want to check below that `lpPermit.owner == msg.sender`?

hardhat.config.ts
```diff
@@ -51,7 +51,7 @@ task("accounts", "Prints the list of accounts", async (args: any, hre: any) => {
 
 const config: HardhatUserConfig = {
   typechain: {
-    outDir: `./typechain`,
+    outDir: `./typechain-types`,
```
mijovic, line 54: After some of @RedFox20 refactors my tests were broken. This solved it

tempus-sdk/utils/ERC20.ts
```diff
@@ -150,4 +160,33 @@ export class ERC20 extends ContractBase implements IERC20 {
...
+    return {
+      owner: owner.address,
+      spender: addressOf(spender),
+      value: this.toBigNum(value),
+      nonce: await this.contract.nonces(owner.address),
+      deadline: deadline
+    };
+  }
+
+  async getDomain(chainId: number): Promise<any> {
```
mijovic, line 184: I think these two should be part of ERC20, let me know wdyt?

mijovic:
Failed tests are `supportsInterface` related, I'll update it once everything is added to interface, don't want to update in each step :)

hardhat.config.ts
RedFox20: Weird, what's up with this?

tempus-sdk/utils/ERC20.ts
```diff
@@ -150,4 +160,33 @@ export class ERC20 extends ContractBase implements IERC20 {
...
+    Permit: [
+      {name: "owner", type: "address"},
+      {name: "spender", type: "address"},
+      {name: "value", type: "uint256"},
+      {name: "nonce", type: "uint256"},
+      {name: "deadline", type: "uint256"}
+    ]
+  };
+
+  async getPermitStruct(owner: SignerWithAddress, spender: Addressable, value: Numberish, deadline: BigNumberish): Promise<PermitStruct> {
```
RedFox20, line 174: You can use `owner: Addressable`

tempus-sdk/tempus/TempusController.ts
```diff
@@ -209,23 +211,43 @@ export class TempusController extends ContractBase {
 
   async exitAmmGivenAmountsOutAndEarlyRedeem(
     pool: PoolTestFixture,
-    user: SignerOrAddress,
+    user: SignerWithAddress,
```
RedFox20, line 214: You can use `Addressable` here as well. Actually it might conflict with my PR...

tempus-sdk/utils/ERC20.ts
```diff
@@ -150,4 +160,33 @@ export class ERC20 extends ContractBase implements IERC20 {
       value: this.toBigNum(amount)
     });
   }
+
+  permitTypes = {
```
RedFox20, line 164: This should probably be static

tempus-sdk/utils/ERC20.ts
RedFox20: I think it's fine. But the return type should probably be removed so that TypeScript can deduce the object fields. Right now the `Promise<any>` is doing type erasure

tempus-sdk/utils/ERC20.ts
```diff
@@ -1,8 +1,18 @@
-import { Contract, Transaction } from "ethers";
+import { BigNumberish, Contract, Transaction } from "ethers";
 import { Decimal } from "./Decimal";
 import { Numberish } from "./DecimalUtils";
 import { ContractBase, Signer, AbstractSigner, Addressable, addressOf } from "./ContractBase";
 import { IERC20 } from "./IERC20";
+import { keccak256, defaultAbiCoder, toUtf8Bytes, solidityPack } from 'ethers/lib/utils';
+import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/dist/src/signers";
```
RedFox20, line 7: Lets not import this, there was a reason I made type aliases in ContractBase -- to make it easier to add/remove types in the future

contracts/TempusController.sol
yuval-weiss: I'd say yeah, let's add a `lpPermit.owner == msg.sender` and also a `lpPermit.spender == address(this)` for good measure

contracts/ITempusController.sol
```diff
@@ -4,6 +4,17 @@ pragma solidity 0.8.10;
 import "./amm/ITempusAMM.sol";
 import "./ITempusPool.sol";
 
+/// Parameters for ERC20Permit.permit call
+struct ERC20PermitSignature {
```
yuval-weiss, line 8: maybe we should move this to a separate interface or library so that this struct can also be used in the TempusAmm

contracts/TempusController.sol
yuval-weiss: actually, I think we can just remove `owner` and `spender` from `ERC20PermitSignature` and just do:
```
tempusAMM.permit(
                msg.sender, 
                address(this), 
                lpPermit.value, 
                lpPermit.deadline, 
                lpPermit.v, 
                lpPermit.r, 
                lpPermit.s
            );
```

If in the future we allow interacting with the controller using a `from` param other than `msg.sender`, we can add `owner` back to the struct

contracts/ITempusController.sol
axic: Isn't this part of `draft-ERC20Permit` in OZ?

axic: Apparently no: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/draft-IERC20Permit.sol

contracts/TempusController.sol
axic: I wonder if it should be `ERC20PermitSignature[] calldata permits` in all the relevant functions, and just apply all the supplied permits. That way it is flexible.

tempus-sdk/utils/ERC20.ts
mijovic: These two are direct dependencies of the contract ser
```
value: this.toBigNum(value),
nonce: await this.contract.nonces(owner.address),
```

contracts/TempusController.sol
mijovic: > 

How to know which permit is used for which token?

tempus-sdk/utils/ERC20.ts
yuval-weiss: I think we should move these 2 functions out of this file as it's `ERC20Permit` that would not work with any `ERC20` token.
I think we should either:
1. Implement @pashov 's suggestion of adding a `PermitUtils`. We can just pass the `ERC20` instance as a parameter to the library and read `this.contract.nonces` and whatever else is needed.
2. Create an `ERC20Permit` class that inherits from `ERC20` and then move both of these functions there. That would cause some typing issues because currently the TempusController tests use the pool test fixture which exposes tokens as `ERC20` instances (and not `ERC20Permit`) - an explicit typecast (`erc20TokenInstance as ERC20Permit`) would work but it is a bit hacky.

tempus-sdk/tempus/TempusController.ts
mijovic: It needs to be a `Signer` as it signs one message

pashov (approved):

tempus-sdk/tempus/TempusController.ts
```diff
@@ -209,23 +211,43 @@ export class TempusController extends ContractBase {
...
     principalsLp: Numberish,
     yieldsLp: Numberish,
     toBackingToken: boolean
   ): Promise<Transaction> {
     const amm = pool.amm, t = pool.tempus;
-    await amm.contract.connect(user).approve(this.address, amm.contract.balanceOf(addressOf(user)));
+    
+    const value = await amm.balanceOf(user);
+
+    const permitData = await amm.getPermitStruct(user, this, value, "100000000000");
```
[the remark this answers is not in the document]
axic: I mean the entire data structure and signature splitting should be a helper. Fine to be in ERC20. But currently this is split between ERC20 and TempusController.

contracts/ITempusController.sol
axic: I'd also agree to move this into its own interface file, to be reused by other contracts.

RedFox20: We should move this into a Library so it can be reused like this:
```ts
/// Parameters for ERC20Permit.permit call
struct ERC20PermitSignature {
    address token;
    uint256 value;
    uint256 deadline;
    uint8 v;
    bytes32 r;
    bytes32 s;
}

library PermitHelper {
    function applyPermit(ERC20PermitSignature calldata p, address owner, address spender) internal {
        IERC20Permit(p.token).permit(owner, spender, p.value, p.deadline, p.v, p.r, p.s);
    }
}
```

And this will keep our business logic much cleaner:
```ts
using PermitHelper for ERC20PermitSignature;
...
        for (uint256 i = 0; i < erc20Permits.length; i++) {
            erc20Permits[i].applyPermit(msg.sender, address(this);
        }
```

tempus-sdk/utils/ERC20.ts
```diff
@@ -150,4 +158,33 @@ export class ERC20 extends ContractBase implements IERC20 {
...
+    Permit: [
+      {name: "owner", type: "address"},
+      {name: "spender", type: "address"},
+      {name: "value", type: "uint256"},
+      {name: "nonce", type: "uint256"},
+      {name: "deadline", type: "uint256"}
+    ]
+  };
+
+  async getPermitStruct(owner: Addressable, spender: Addressable, value: Numberish, deadline: BigNumberish): Promise<PermitStruct> {
```
RedFox20, line 172: I think we shouldn't take dependency on `BigNumberish` anymore and just try using `number` here, it will work with timestamps

tempus-sdk/tempus/TempusController.ts
```diff
@@ -209,23 +211,43 @@ export class TempusController extends ContractBase {
...
+    ));
+
+    const lpPermit = {
+      token: amm.address,
+      value: permitData.value,
+      deadline: permitData.deadline,
+      v: signature.v,
+      r: signature.r,
+      s: signature.s
+    };
```
RedFox20, line 240: Creating this struct should be some sort of a new utility so we can reuse it in SDK utils

tempus-sdk/utils/ERC20.ts
```diff
@@ -1,9 +1,17 @@
...
+import { BigNumberish, Contract, Transaction } from "ethers";
 import { Decimal } from "./Decimal";
 import { Numberish } from "./DecimalUtils";
 import { ContractBase, Signer, Addressable, addressOf } from "./ContractBase";
 import { IERC20 } from "./IERC20";
 
+interface PermitStruct {
+  owner: string;
+  spender: string;
+  value: BigNumberish;
```
RedFox20, line 10: We should use `Numberish` here instead. It will give the same effect, but we don't need to depend on ethers BigNumberish. I want us to move away from that BigNumber library completely

tempus-sdk/utils/ERC20.ts
RedFox20: WDYM? `toBigNum` is a utility in `DecimalConvertible` base class which stores the `decimals` variable. Makes number conversion very convenient. They will be named `toContractNum` and `fromContractNum` soon, like in Starknet utils

contracts/ITempusController.sol
mijovic: Implemented @RedFox20's suggestion, it makes the code pretty simple

tempus-sdk/tempus/TempusController.ts
```diff
@@ -250,14 +282,54 @@ export class TempusController extends ContractBase {
...
+      [],
+      amm.toBigNum(lpTokens),
+      amm.principalShare.toBigNum(principals),
+      amm.yieldShare.toBigNum(yields),
+      {
+        minPrincipalsStaked: 0,
+        minYieldsStaked: 0,
+        maxLeftoverShares: t.principalShare.toBigNum(maxLeftoverShares),
+        yieldsRate: amm.principalShare.toBigNum(yieldsRate),
+        maxSlippage: toWei(maxSlippage)
```
RedFox20, line 294: Is it `Wei` ? Or should we use `amm.toBigNum(maxSlippage)` here?

tempus-sdk/tempus/TempusController.ts
```diff
@@ -209,28 +210,59 @@ export class TempusController extends ContractBase {
...
+    principals: Numberish,
+    yields: Numberish,
+    principalsLp: Numberish,
+    yieldsLp: Numberish,
+    toBackingToken: boolean,
+    deadline: Date = new Date(8640000000000000) /// default is 9/12/275760 (no deadline)
+  ): Promise<Transaction> {
+    const amm = pool.amm, t = pool.tempus;
+    
+    const lpPermit = await constructPermit(amm, user, this, await amm.balanceOf(user), deadline);
```
RedFox20, line 223: So this is the new Approve through permits 👍🏻

mijovic: Yes ser

contracts/utils/PermitHelper.sol
```diff
@@ -0,0 +1,24 @@
...
+}
+
+library PermitHelper {
+    function applyPermit(
+        ERC20PermitSignature calldata p,
+        address owner,
+        address spender
+    ) internal {
+        p.token.permit(owner, spender, p.value, p.deadline, p.v, p.r, p.s);
+    }
```
yuval-weiss, line 23:
```suggestion
    }
    
    function applyPermits(
        ERC20PermitSignature[] calldata permits,
        address owner,
        address spender
    ) internal {
        for (uint256 i = 0; i < permits.length; i++) {
              applyPermit(permits, msg.sender, address(this));
          }
    }
```

and then from the TempusController functions that support permits, we can save the code duplication of running a for-loop over the provided permits

test/TempusController.test.ts
```diff
@@ -592,6 +648,88 @@ describeForEachPool("TempusController", (testPool:PoolTestFixture) =>
         expect(+await testPool.asset.balanceOf(user1)).to.be.within(1199000, 1200000);
       }
     });
+
+    it("Complete exit to yield bearing with approvals", async () =>
```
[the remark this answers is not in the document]
mijovic: Updated, also removed duplication in TempusController.ts

RedFox20 (approved):
Looks good to me 👍🏻

PR #650, opened 2022-06-09
sdk: implement Decimal functions to remove BigNumber from contracts

RedFox20:


mijovic (dismissed):

mijovic:
Tests are failing

tempus-sdk/utils/Decimal.ts
```diff
@@ -180,6 +203,10 @@ export class Decimal {
     return this.int <= this.toScaledBigInt(x);
   }
 
+  public isZero(): boolean {
+    return this.int === ZERO;
```
axic, line 207:
```suggestion
    return this.int === 0n;
```

Does the extra const provides any benefit?

test/Decimal.test.ts
```diff
@@ -333,16 +340,17 @@ describeNonPool("Decimal", () =>
...
-    function bn18(x:Numberish): BigNumber {
+    function bn18(x:Numberish): bigint {
       const parts = x.toString().split('.');
-      const whole = BigNumber.from(parts[0]);
+      const whole = BigInt(parts[0]);
       if (parts.length === 1) // it was an integer, return it as-is
         return whole;
       // it's a decimal fraction, scale it up:
-      const sign = BigNumber.from(whole.isNegative() ? '-1' : '1');
+      const sign = BigInt(whole >= 0 ? '1' : '-1');
```
axic, line 349: I think it maybe safer to use the same type?

```suggestion
      const sign = BigInt(whole >= 0n ? '1' : '-1');
```

test/Decimal.test.ts
```diff
@@ -333,16 +340,17 @@ describeNonPool("Decimal", () =>
...
       const fractString = parts[1].padEnd(18, '0'); // pad 04 to 040000000000000000
-      const fract = BigNumber.from(fractString);
-      const scaledResult = whole.abs().mul('1000000000000000000').add(fract).mul(sign);
+      const fract = BigInt(fractString);
+      const absWhole = whole >= 0 ? whole : -whole;
```
axic, line 352: Same here:
```suggestion
      const absWhole = whole >= 0n ? whole : -whole;
```

test/Decimal.test.ts
```diff
@@ -333,16 +340,17 @@ describeNonPool("Decimal", () =>
...
+      const scaledResult = (absWhole * BigInt('1000000000000000000') + fract) * sign;
```
axic, line 353: Why not also use the `n` suffixed literal here?

tempus-sdk/utils/DecimalUtils.ts
```diff
@@ -14,84 +14,76 @@ export const MAX_NUMBER_DIGITS = 17;
...
- * Converts any input number into an ethers.BigNumber.
+ * Converts any input number into a BigInt.
  * Regular numbers are simply truncated to integer.
- * To create scaled BigNumbers, pass a Decimal to this function.
+ * To create scaled BigInt, pass a Decimal to this function.
  * Example: bn(decimal(1.0, 18)) -> 1000000000000000000
  * 
  * @param number Any number-like value
  */
 export function bn(number:Numberish): BigNumber {
```
axic, line 27: Will this be moved to `bigint` too?

tempus-sdk/utils/ERC20Ether.ts
```diff
@@ -34,7 +34,7 @@ export class ERC20Ether extends DecimalConvertible implements IERC20 {
    */
   async balanceOf(account:Addressable): Promise<Decimal> {
     const balance = await ethers.provider.getBalance(addressOf(account));
```
axic, line 36: What is the type of this?

tempus-sdk/utils/DecimalUtils.ts
```diff
@@ -14,84 +14,76 @@ export const MAX_NUMBER_DIGITS = 17;
...
+ * Converts a decimal number into a scaled bigint
  * @example let wei = parseDecimal("0.000001", 18);
  * @param decimal Decimal such as 1.25 or "12.1234777777"
  * @param decimalBase Base precision of the decimal, for wei=18, for ray=27 
- * @returns BigNumber for use in solidity contracts
+ * @returns Scaled bigint for use in solidity contracts
  */
-export function parseDecimal(decimal:Numberish, decimalBase:number): BigNumber {
+export function parseDecimal(decimal:Numberish, decimalBase:number): bigint {
   // need this special case to support MAX_UINT256, ignoring decimalBase
```
axic, line 45: Is this still a valid problem?

tempus-sdk/utils/DecimalUtils.ts
RedFox20: I went through it and decided it's way out of scope, so I had to revert from this:
```ts
export function bn(number:Numberish): BigInt {
  if (typeof(number) === "bigint")
    return number;
  if (number instanceof Decimal)
    return number.toBigInt();
  // truncate decimal part
  const whole = number.toString().split('.')[0];
  return BigInt(whole);
}
```

axic: Because the description above is updated.

tempus-sdk/utils/DecimalUtils.ts
```diff
@@ -1,7 +1,7 @@
 import { BigNumber } from "ethers";
 import { Decimal } from "./Decimal";
 
-export type Numberish = Number | number | string | BigInt | BigNumber | Decimal;
```
axic, line 4: What is the difference between `Number` and `number`, because the latter is now removed?

tempus-sdk/utils/ERC20Ether.ts
RedFox20: Probably a `BigNumber`. I was trying to remove support for it, but I'll have to do a second pass on this in a new PR.

integration-tests/TempusPool.rari.test.ts
```diff
@@ -29,7 +29,7 @@ const setupWithRariWithdrawalFee = async (rariFee:Decimal) => await deployments.
   if (rariFee.valueOf() > 0) {
     await owner.sendTransaction({ from: owner.address, to: rariFundManagerOwner.address, value: toWei(3) });
     // Set Rari's Withdrawal Fee
-    await rariFundManager.connect(rariFundManagerOwner).setWithdrawalFeeRate(rariFee.toBigNumber());
+    await rariFundManager.connect(rariFundManagerOwner).setWithdrawalFeeRate(rariFee.toBigInt());
```
mijovic, line 32: why do we still have Rari pools in repo?

test/Decimal.test.ts
RedFox20: We'll have to upgrade ES in another PR to get that feature

tempus-sdk/utils/Decimal.ts
RedFox20: We'll have to upgrade ES in another PR to get that feature

tempus-sdk/utils/DecimalUtils.ts
RedFox20: I think we only have comparison to `MAX_UINT256` remaining. There is nothing that formats this. So it could be removed, unless we want to keep this as a feature to send MAX_UINT256 to contracts...

tempus-sdk/utils/DecimalUtils.ts
RedFox20: In this case it's just for type checking. If `type X = number;` then passing `Number(1)` to it will fail. But `type X = Number;` uses the interface `Number`, which will work with both `Number ` and `number` values

integration-tests/TempusPool.rari.test.ts
RedFox20: Just in case ™️

tempus-sdk/utils/DecimalUtils.ts
RedFox20: Ah, I'll revert that

PR #658, opened 2022-06-16
Update to Solidity 0.8.15

axic:
Need to measure gas differences. Should consider turning on IR.


RedFox20:
> Need to measure gas differences. Should consider turning on IR.

Is there a guaranteed improvement compared to previous version? If so, we should consider merging it

pashov:
Closing for the following reasons:
1. Solidity 0.8.17 is out already
2. The build was failing on this PR
3. There are merge conflicts on this PR

PR #669, opened 2022-06-28
Use tempus utils as a package.

mijovic:


mijovic (commented):
I triple checked all changes, all of them are either deleting file or changing import statements

contracts/ITempusPool.sol
```diff
@@ -2,9 +2,9 @@
 pragma solidity 0.8.10;
 
 import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+import "@tempus-labs/tempus-utils/contracts/utils/IOwnable.sol";
```
RedFox20, line 5: Taking inspiration from @openzeppelin, then this should be more like:
```ts
import "@tempus-labs/contracts/utils/IOwnable.sol";
```
Is there a limitation with how NPM imports are set up?

test/incentivization/LeverageIncentivization.test.ts
```diff
@@ -2,11 +2,11 @@ import { expect } from "chai";
 import { Contract } from "ethers";
 import { ethers, waffle } from "hardhat";
 import { MockContract } from "@ethereum-waffle/mock-contract";
-import { ContractBase, Signer } from "@tempus-sdk/utils/ContractBase";
-import { blockTimestamp, expectRevert, impersonateAccount, setBalanceOf, setEvmTime } from "@tempus-sdk/utils/Utils";
-import { decimal } from "@tempus-sdk/utils/Decimal";
-import { Numberish, parseDecimal } from "@tempus-sdk/utils/DecimalUtils";
-import { ERC20 } from "@tempus-sdk/utils/ERC20";
+import { ContractBase, Signer } from "@tempus-labs/tempus-utils/build/utils/ContractBase";
```
RedFox20, line 5: I recommend we try this naming instead, because this is really hard to change in the future. Better to make it nice right away.
```ts
import { } from "@tempus-labs/utils/ContractBase";
```
It can be solved in `tsconfig.json` to handle `"paths"` aliases.
```json
{
  "compilerOptions": {
    // Need this for VS Code to properly detect external library type definitions
    "baseUrl": ".",
    "sourceMap": true,
    "paths": {
      "@tempus-labs/*": ["@tempus-labs/tempus-utils/build*"],
    }
  }
}
```

yarn.lock
```diff
@@ -781,6 +781,13 @@
   dependencies:
     defer-to-connect "^1.0.1"
 
+"@tempus-labs/tempus-utils@^1.0.6":
+  version "1.0.6"
+  resolved "https://registry.yarnpkg.com/@tempus-labs/tempus-utils/-/tempus-utils-1.0.6.tgz#e12d11be95dc555d5b6dde54420199f8eb1a39db"
+  integrity sha512-AqxCqp/I3g+n7xsn4rq2T5i+fv1eivhOTzLIRjEill8gngGMAu0/hYuRkXVDrBYBr25ZaT3OGB9bR2foQw2fgw==
+  dependencies:
+    "@openzeppelin/contracts" "^4.1.0"
```
RedFox20, line 789: We should probably use the same convention as `@openzeppelin` uses here. `@tempus-labs/utils`

test/token/TempusToken.test.ts
```diff
@@ -1,7 +1,7 @@
 import { ethers } from "hardhat";
 import { expect } from "chai";
-import { blockTimestamp, expectRevert, increaseTime } from "@tempus-sdk/utils/Utils";
-import { Signer } from "@tempus-sdk/utils/ContractBase";
+import { blockTimestamp, expectRevert, increaseTime } from "@tempus-labs/tempus-utils/build/utils/Utils";
```
RedFox20, line 3: By using the `build/` folder here, probably something not set up correctly in the NPM package

mijovic: that is how it is used. You need to compile each ts package to get output which can be imported from external projects. However, I changed out dir to dist. Should look better now.

contracts/ITempusPool.sol
mijovic: it is `@org-name/package-name`
Since these are general utils and not only contracts we can't do that right now.

mijovic: What we can do is to setup two packages from utils, then it would be able to work like this

mijovic: Solved with 2 separate packages, one for contracts, the other one for ts utils

test/incentivization/LeverageIncentivization.test.ts
mijovic: I think no need for this now, as it looks much better

yarn.lock
```diff
@@ -781,6 +781,18 @@
   dependencies:
     defer-to-connect "^1.0.1"
 
+"@tempus-labs/contracts@^1.0.0":
+  version "1.0.0"
+  resolved "https://registry.yarnpkg.com/@tempus-labs/contracts/-/contracts-1.0.0.tgz#486d3f44f2bb8d80a4cb414a5274d7c4aca7a2cc"
+  integrity sha512-RL7WUF2dn4E0jQzP0vn9NbXnFHSQa27TkDOd/Egw20nn8x979YtwSIqoyqUbrItbfVxBfEbl/Bpx84z/Ot3lyg==
+  dependencies:
+    "@openzeppelin/contracts" "^4.1.0"
```
pashov, line 789: Update OZ dependency to latest version

mijovic: I would do this in separate PRs. This comes from package.json for tempus-protocol

pashov (approved):

PR #673, opened 2022-08-11
Add referral support

### LINKED ISSUE #612: Add referral support to TempusPools

pashov:
Fixes #612


contracts/PositionManager.sol
```diff
@@ -66,7 +66,8 @@ contract PositionManager is IPositionManager, ERC721, ReentrancyGuard {
                 tokenAmountToDeposit,
                 params.isBackingToken,
                 backingTokenONE.divfV(params.worstAcceptableCapitalsRate, backingTokenONE),
-                params.deadline
+                params.deadline,
+                address(0)
```
mijovic, line 70: maybe adding an option for referral via function parameter? This would help people to pass it even if it is done via PositionsManager?

contracts/TempusController.sol
```diff
@@ -285,9 +290,10 @@ contract TempusController is ITempusController, ReentrancyGuard, Ownable, ERC165
         ITempusAMM tempusAMM,
         ITempusPool tempusPool,
         uint256 tokenAmount,
-        bool isBackingToken
+        bool isBackingToken,
+        address referral
```
mijovic, line 294: I think we need this for redemptions

mijovic: Never mind, github showed weird diff, all good

mijovic (approved):

PR #676, opened 2022-08-16
Add permit support to depositAndFix and depositAndLeverage

### LINKED ISSUE #672: [Controller] Implement Permit for depositAndFix, depositAndLeverage

pashov:
Fixes #672


contracts/ITempusController.sol
```diff
@@ -168,6 +168,9 @@ interface ITempusController {
     ///      See https://docs.balancer.fi/developers/guides/single-swaps#swap-overview
     /// @param tempusAMM Tempus AMM to use to swap TYS for TPS
     /// @param tempusPool The Tempus Pool to which tokens will be deposited
+    /// @param erc20Permits Permit signatures for spending tokens that needs approvals
+    ///                     This method needs approval for LP tokens
```
mijovic, line 172: I think this line is not true. LP token doesn't have anything to do with this method. I think you need approvals for `backing`/`yieldBearingTokens`

contracts/ITempusController.sol
```diff
@@ -187,6 +191,9 @@ interface ITempusController {
 
     /// @dev Atomically deposits YBT/BT to TempusPool and swaps Capitals for Yields to get leveraged exposure to yield
     /// @param tempusPool TempusPool to be used for depositing YBT/BT
+    /// @param erc20Permits Permit signatures for spending tokens that needs approvals
+    ///                     This method needs approval for LP tokens
```
mijovic, line 195: Same here

test/TempusController.test.ts
```diff
@@ -72,7 +72,7 @@ describeForEachPool("TempusController", (testPool:PoolTestFixture) =>
       expect(await controller.supportsInterface("0x3c3dbb51")).to.be.false;
   
       // should support ITempusController interface
-      expect(await controller.supportsInterface("0x9c03fa10")).to.be.true;
+      // expect(await controller.supportsInterface("0x9c03fa10")).to.be.true;
```
mijovic, line 75: we need to fix this test

pashov:
My approach will be to change the yield bearing tokens for each protocol defined in the `mocks/` directory to support ERC20Permit and also change  the deployed "asset" or underlying token in the code, that is usually an instance of `ERC20FixedSupply` to be an instance of my added mock `MockERC20Permit`.

I expect I will hit some troubles with Lido, because the mock token for it does not inherit from ERC20, so if I want it to inherit from ERC20Permit it might not really work with Solidity

test/TempusController.test.ts
mijovic: just bumping it up so we do not forget

test/TempusController.test.ts
```diff
@@ -152,79 +152,110 @@ describeForEachPool("TempusController", (testPool:PoolTestFixture) =>
 
   describe("depositAndFix", () =>
   {
-    it("verifies tx reverts if provided minimum TYS rate requirement is not met", async () =>
+    it.only("verifies tx reverts if provided minimum TYS rate requirement is not met", async () =>
```
mijovic, line 155:
```suggestion
    it("verifies tx reverts if provided minimum TYS rate requirement is not met", async () =>
```

test/TempusController.test.ts
```diff
@@ -152,79 +152,110 @@ describeForEachPool("TempusController", (testPool:PoolTestFixture) =>
...
     });
 
     it("verifies depositing YBT succeeds if provided minimum TYS rate requirement is met", async () =>
     {
-      await initAMM(user1, /*ybtDeposit*/2000, /*principals*/200, /*yields*/2000); // 10% rate
-      const minTYSRate = "0.097"; // 9.7% (fee + slippage)
-      await controller.depositAndFix(testPool, user2, 5.456789, false, minTYSRate); 
-      await expectValidState();
+      await shouldSucceedYieldBearingTokenMinimumTysRate(true);
+      // await shouldSucceedYieldBearingTokenMinimumTysRate(false);
```
mijovic, line 172: maybe remove this line?

test/TempusController.test.ts
```diff
@@ -152,79 +152,110 @@ describeForEachPool("TempusController", (testPool:PoolTestFixture) =>
...
     it("verifies depositing BT succeeds if provided minimum TYS rate requirement is met", async () =>
     {
-      await initAMM(user1, /*ybtDeposit*/2000, /*principals*/20, /*yields*/200); // 10% rate
-      const minTYSRate = "0.097"; // 9.7% (fee + slippage)
-      const amount = 5.456789;
-      const ethAmount = testPool.type == PoolType.Lido ? amount : 0;
-      await controller.depositAndFix(testPool, user2, 5.456789, true, minTYSRate, ethAmount); 
-      await expectValidState();
+      await shouldSucceedBackingTokenWithMinimumTysRate(true);
+      // await shouldSucceedBackingTokenWithMinimumTysRate(false);
```
mijovic, line 188: why no separate test to run with false?

mijovic (commented):
A bunch of tests with `false` for use approvals are commented out. Otherwise it looks good

pashov:
> A bunch of tests with `false` for use approvals are commented out. Otherwise it looks good

All of those are testing the new permit functionality, but they still don't work.

My progress on this is that I made it so that both the `asset` and `yieldBearing` tokens for every test pool are now ERC20Permit compatible as I mentioned in my comment above.
Now the tests are failing with `ERC20: transfer amount exceeds balance` but it looks like it's not in a place where signature permits are used.

I think I might have a problem with the token decimals, maybe the allowance is not zero but just less than what is needed.

test/TempusController.test.ts
pashov: Fixed the test

test/TempusController.test.ts
pashov: Fixed the test

mijovic (approved):

